PR #2, opened 2021-06-21
rename repo to qt-extras

tlambert03:


jni (approved):
I'm approving, but just to throw another one out there that just occurred to me: superqt. What do you think?

tlambert03:
More adorable, slightly less apropos :)
I could be ok with it

jni:
> slightly less apropos :)

I like all of these for this library 😂 

Screen Shot 2021-06-23 at 12 08 12 pm

tlambert03:
gonna go with `superqt`

PR #7, opened 2021-07-01
[wip] pure python qt signal/slot implementation

tlambert03:
playing around with this for now... see https://github.com/napari/superqt/issues/6 for context.

This is a pure python re-implementation of the Qt signal/slot mechanism.   similar to the deprecated https://github.com/dgovil/PySignal repo... with some updates.

for me, this has some advantages over the `event.py` module we use in napari:

- callback signature support: you could know at the moment of calling `.connect` whether the callback is compatible.
- defining callbacks is clearer when you can list the parameters in the callback signature.  In the current napari system all callbacks take a single event object, and it's not immediately clear (and even harder to "type") what properties that event object should have.  (i.e. `event.value`?) (This is related to https://github.com/napari/magicgui/pull/182)
- familiar API: lots of people already know the (py)Qt signal/slot API.  so this would provide a pure-python implementation (for things like napari where we don't necessary want the entire callback system to be tied to Qt), but with a more or less cross-compatible API

note: this does *not* currently have the thread safety that the Qt signals system does (that would be a bigger lift).  but the same could be said for the existing napari/vispy system.


alisterburt:
> familiar API: lots of people already know the (py)Qt signal/slot API. so this would provide a pure-python implementation (for things like napari where we don't necessary want the entire callback system to be tied to Qt), but with a more or less cross-compatible API

This point is the most important for me - I've definitely stumbled over the coexisting events systems in napari a couple of times! Awesome that you're pushing this

superqt/signal/_signal.py
```diff
@@ -0,0 +1,181 @@
...
+        if instance is None:
+            return self
+        d = self._signal_instances.setdefault(self, weakref.WeakKeyDictionary())
+        return d.setdefault(instance, SignalInstance(self.signatures, instance))
+
+
+class SignalInstance:
+    def __init__(
+        self, signatures: tuple[Signature, ...] = (), instance: Any = None
+    ) -> None:
```
alisterburt, line 84: is the difference between `Signal` and `SignalInstance` that `Signal` is the type of a signal and the other is a specific instance of a given signal?

tlambert03: yeah, pretty much.  it's confusing... but interesting! and get's to the heart of python's [descriptor protocol](https://docs.python.org/3/howto/descriptor.html) (the same thing behind the `@property` decorator).  but, consider this code with standard Qt signals:

```py
In [1]: from qtpy.QtCore import QObject, Signal

In [2]: class MyObj(QObject):
   ...:     changed = Signal() 
   ...:

In [3]: my_obj = MyObj()

In [4]: type(MyObj.changed)
Out[4]: <class 'PySide2.QtCore.Signal'>

In [5]: type(my_obj.changed)
Out[5]: <class 'PySide2.QtCore.SignalInstance'>
```

`changed` is a class attribute (i.e. it doesn't know about self).  When you access the `changed` attribute on the `MyObj` class, you get an instance of the `Signal` type... like you'd expect. 

`Signal` however, implements the descriptor protocol (it provides a [`__get__` method](https://docs.python.org/3/reference/datamodel.html#object.__get__)), which lets you detect whether the attribute is being accessed on the Class itself, or an instance of the class (`my_obj`).  If we are being accessed on an instance, we return a new object that handles all of the callback connections to _that specific_ instance.

This is what allows different instances of MyObj to have their own callbacks connected:

```py
# two instances
my_obj = MyObj()
my_other_obj = MyObj()

# one connected
my_obj.changed.connect(do_something)

# other emits
my_other_obj.changed.emit()

# nothing happens... because the my_other_obj.changed SignalInstance has no callbacks
```

tlambert03:
this is coming along nicely.  with good validation and type/signature safety when connecting a slot.  It's also performing comparably too (napari events is faster on `.connect` but `superqt` does a lot of validation during connection, and is faster on object creation, and slightly faster where it really counts: on emission).  anyway, negligible differences. but more safety here.

```
------------------------------------------- benchmark 'connect': 3 tests ------------------------------------------
Name (time in ns)                           Min                        Max                   Mean            Rounds
-------------------------------------------------------------------------------------------------------------------
test_connect_napari_bench              613.2500 (1.0)          34,523.7500 (1.0)         682.1718 (1.0)       78193
test_connect_superqt_bench           7,829.0000 (12.77)    28,176,299.0000 (816.14)   10,361.8767 (15.19)     23657
test_connect_superqt_bench_typed     8,961.0000 (14.61)       383,588.0000 (11.11)    10,242.6791 (15.01)     39159
-------------------------------------------------------------------------------------------------------------------

-------------------------------------- benchmark 'create': 2 tests --------------------------------------
Name (time in ns)                    Min                     Max                   Mean            Rounds
---------------------------------------------------------------------------------------------------------
test_create_superqt_bench       164.6100 (1.0)        2,289.3500 (1.0)         175.1490 (1.0)       48741
test_create_napari_bench      9,435.0000 (57.32)    152,593.0000 (66.65)    10,253.4639 (58.54)     10766
---------------------------------------------------------------------------------------------------------

------------------------------- benchmark 'emit': 2 tests -------------------------------
Name (time in us)              Min                Max              Mean            Rounds
-----------------------------------------------------------------------------------------
test_emit_superqt_bench     3.0470 (1.0)      34.9530 (1.0)      3.2908 (1.0)       81136
test_emit_napari_bench      3.3810 (1.11)     41.7790 (1.20)     3.7392 (1.14)      44739
-----------------------------------------------------------------------------------------
```

tlambert03:
after some thought... I'm not sure this is the right library for this.  `superqt` is really designed to _extend_ and be used with Qt, where as this PR is designed to _mimic_ (but be useable without) Qt.  I'm going to put it in its own repo for now: https://github.com/tlambert03/psygnal

we can perhaps use it here (any maybe napari someday?).  and I'll probably use it in magicgui since i think the end-user interface is much easier than our current one (where the end-user needs to accept a single mysterious `Event` argument and find some attribute that may or may not be named `"value"` :))

PR #13, opened 2021-07-23
Enum ComboBox implementation

Czaki:
This PR add Enum ComboBox to superqt. 
Enum is quite popular method of limit selection of options so it will be nice to simplify using it from interface.


superqt/combobox/_enum_combobox.py
```diff
@@ -0,0 +1,84 @@
...
+        super().addItems(list(map(_get_name, enum.__members__.values())))
+
+    def enum(self) -> Optional[Type[EnumType]]:
+        return self._enum
+
+    def clear(self):
+        self._enum = None
+        super().clear()
+
+    def currentValue(self) -> EnumType:
```
tlambert03, line 50: curious whether you considered and rejected using `currentData`?  Do you think it's best to leave that for user usage?

I'm just thinking about alternatives to "value" ... it doesn't immediately imply to me that we're getting the enum member...

superqt/combobox/_test/enum_comb_box.py
```diff
@@ -0,0 +1,102 @@
+from enum import Enum
```
tlambert03, line 1: could you rename this file `test_enum_comb_box.py`, and rename the directory `_tests` (plural) to match the rest of the repo?

superqt/combobox/_enum_combobox.py
```diff
@@ -0,0 +1,84 @@
...
+        return list(self._enum.__members__.values())[self.currentIndex()]
+
+    def setValue(self, value: Union[EnumType, int]):
+        """Set value with Enum or int"""
+        if not isinstance(value, (Enum, int)):
+            return
+        if isinstance(value, Enum):
+            self.setCurrentText(_get_name(value))
+        else:
+            self.setCurrentIndex(value)
```
tlambert03, line 63: maybe just accept Enum, and leave setCurrentIndex as the "one right way" to set the index?  what do you think?

```suggestion
    def setValue(self, value: EnumType) -> None:
        """Set value with Enum."""
        if not isinstance(value, Enum):
            raise TypeError(f'setValue(self, Enum): argument 1 has unexpected type {type(value).__name__!r}')
        self.setCurrentText(_get_name(value))
```

superqt/combobox/_test/enum_comb_box.py
```diff
@@ -0,0 +1,102 @@
...
+    qtbot.addWidget(enum)
+    with pytest.raises(RuntimeError):
+        enum.addItem("aaa")
+    with pytest.raises(RuntimeError):
+        enum.addItems( ["aaa", "bbb"])
+    with pytest.raises(RuntimeError):
+        enum.insertItem(0, "aaa")
+    with pytest.raises(RuntimeError):
+        enum.insertItems(0, ["aaa", "bbb"])
+    assert enum.count() == 3
```
tlambert03, line 102: newline?

tlambert03:
```suggestion
    assert enum.count() == 3

```

superqt/combobox/_enum_combobox.py
```diff
@@ -0,0 +1,84 @@
+from typing import Optional, Type, TypeVar, Union
+from enum import Enum
+
+from ..qtcompat.QtWidgets import QComboBox
+from ..qtcompat.QtCore import Signal, PYQT5
+
+EnumType = TypeVar("EnumType", bound=Enum)
+enum_type = Enum if PYQT5 else object
```
tlambert03, line 8: interesting... does PyQt handle python enums better than PySide2?

superqt/combobox/_enum_combobox.py
```diff
@@ -0,0 +1,84 @@
...
+        name = str(enum_value)
+    else:
+        name = enum_value.name.replace("_", " ")
+    return name
+
+
+class EnumComboBox(QComboBox):
+    """
+    This is implementation of variant of EnumComboBox for select enum value.
+    If enum class does not provide own __str__ implementation then human readable name is provide using name of item
```
tlambert03, line 23: break line

superqt/combobox/_enum_combobox.py
```diff
@@ -0,0 +1,84 @@
...
+        # check if function was overloaded
+        name = str(enum_value)
+    else:
+        name = enum_value.name.replace("_", " ")
+    return name
+
+
+class EnumComboBox(QComboBox):
+    """
+    This is implementation of variant of EnumComboBox for select enum value.
```
tlambert03, line 22: place comment on first line, with [empty line between it and description](https://www.python.org/dev/peps/pep-0257/#id17).

superqt/combobox/_enum_combobox.py
```diff
@@ -0,0 +1,84 @@
...
+        return self._enum
+
+    def clear(self):
+        self._enum = None
+        super().clear()
+
+    def currentValue(self) -> EnumType:
+        """current value as Enum member"""
+        if self._enum is None:
+            raise RuntimeError("Enum value is None")
```
tlambert03, line 53: thoughts on just returning `None` rather than raising an Error?

superqt/combobox/_enum_combobox.py
```diff
@@ -0,0 +1,84 @@
...
+            return
+        if isinstance(value, Enum):
+            self.setCurrentText(_get_name(value))
+        else:
+            self.setCurrentIndex(value)
+
+    def _emit_signal(self):
+        try:
+            self.currentValueChanged.emit(self.currentValue())
+        except RuntimeError:
```
tlambert03, line 68: is this just to catch the `if self._enum is None:` above?  or are there other cases where this will err?

tlambert03 (commented):
love it. very qt-like :)
left some nit picks and questions...

superqt/combobox/_enum_combobox.py
Czaki: I think that in the future this widget should also support Optional annotation (I do not know who is the best option to check Optional for now). 

So I prefer exception to fast inform User that he does not properly initialize widget and use option which is distinguishable from None selected.

superqt/combobox/_enum_combobox.py
Czaki: I don't think about `currentData`, I miss it.
But `currentData` should return QVariant so return Enum from `currentData` may break Qt API compatibility?

superqt/combobox/_enum_combobox.py
Czaki: In the past, yes. Now I do not know. I need to find in which scenario it was broken.

superqt/combobox/_enum_combobox.py
Czaki: I do not understand. Maybe I should write docstring from scratch?

superqt/combobox/_enum_combobox.py
```diff
@@ -0,0 +1,84 @@
...
+        name = enum_value.name.replace("_", " ")
+    return name
+
+
+class EnumComboBox(QComboBox):
+    """
+    This is implementation of variant of EnumComboBox for select enum value.
+    If enum class does not provide own __str__ implementation then human readable name is provide using name of item
+    replacing underscores with spaces.
+    """
```
tlambert03:
```suggestion
    """ComboBox presenting options from a python Enum.

    If the Enum class does not implement `__str__` then a human readable name
    is created from the name of the enum member, replacing underscores with spaces.
    """
```
sorry, i wasn't very clear there :)

superqt/combobox/_enum_combobox.py
tlambert03: in my experience qvariant has worked pretty well with most python types... however, after playing with it a bit, I think I've changed my mind: I think we should leave `currentData` for user-space.

I still don't love `currentValue` though... so what do you think about the following?

`setEnumClass` -> clear combo and set enum class (replaces current `setEnum`)
`enumClass` -> return the enum class  (replaces current `enum`)

`currentEnum` -> returns the currently selected enum _member_
`setCurrentEnum` -> set the currently selected enum _member_ (and mark as a slot to match [slots](https://doc.qt.io/qt-5/qcombobox.html#public-slots))
change `currentValueChanged` to `currentEnumChanged`

(I'm still thinking this through, so let me know if you don't like it)

superqt/combobox/_enum_combobox.py
tlambert03: crap... sorry, looks like I commented on an old commit.  here was my suggestion

```
    """ComboBox presenting options from a python Enum.

    If the Enum class does not implement `__str__` then a human readable name
    is created from the name of the enum member, replacing underscores with spaces.
    """
```

superqt/combobox/_enum_combobox.py
tlambert03: hm... maybe let's leave that for when we add an optional (null) member?  ... or go ahead and add it here?  I feel like an exception in something like `currentX` is rather unusual in Qt... normally it would give you either None or some sort of null value like `-1`

tlambert03: for instance, here's an "uninitialized" QComboBox... no exceptions raised:
```py
In [12]: from qtpy.QtWidgets import QComboBox

In [13]: q = QComboBox()

In [14]: q.currentIndex()
Out[14]: -1

In [15]: q.currentText()
Out[15]: ''

In [16]: q.currentData()
```

superqt/combobox/_enum_combobox.py
Czaki: changed. I`m not sure if `setCurrentEnum` or `setEnum` should be used.

superqt/combobox/_enum_combobox.py
Czaki: changed

superqt/combobox/_enum_combobox.py
tlambert03: qt seems to use `setCurrentText` and `setCurrentIndex` (instead of `setText` and `setIndex`) ... so that's why I suggested that pattern.

superqt/combobox/_tests/test_enum_comb_box.py
```diff
@@ -0,0 +1,101 @@
...
+    assert enum.count() == 3
+    assert [enum.itemText(i) for i in range(enum.count())] == ["a 1", "b 2", "c 3"]
+
+
+def test_change_value(qtbot):
+    enum = EnumComboBox(enum_class=Enum1)
+    qtbot.addWidget(enum)
+    assert enum.currentEnum() == Enum1.a
+    with qtbot.waitSignal(enum.currentEnumChanged):
+        enum.setCurrentEnum(Enum1.c)
```
tlambert03, line 81:
```suggestion
    with qtbot.waitSignal(
        enum.currentValueChanged, check_params_cb=lambda x: isinstance(x, Enum)
    ):
        enum.setValue(Enum1.c)
```

(since it's a new signal, let's also check the type emitted)

tlambert03 (commented):
this is looking great.  two more small things:

A) can you export this in the top level `__init__`? (i think we should allow `from superqt import ...` for most things, unless it starts getting unruly
B) can you add a small doc at `docs/combobox.md`?  You can use the template at https://github.com/napari/superqt/blob/main/docs/sliders.md#range-slider as an example.  Doesn't need to be much, but just somewhere we can document the new methods.
C) do we want to call it QEnumComboBox?

docs/combobox.md
```diff
@@ -0,0 +1,31 @@
+# ComboBox
+
+
+## Enum Combo Box
+
```
tlambert03, line 5:
```suggestion

`QEnumComboBox` is a variant of [`QComboBox`](https://doc.qt.io/qt-5/qcombobox.html) 
that populates the items in the combobox based on a python `Enum` class.  In addition to all 
of the methods provided by `QComboBox`, this subclass adds the methods 
`enumClass`/`setEnumClass` to get/set the current `Enum` class represented by the combobox, 
and `currentEnum`/`setCurrentEnum` to get/set the current `Enum` member in the combobox

```

tlambert03:
This looks pretty good to me @Czaki!  Anything else that needs to be done?

tox.ini
```diff
@@ -56,4 +56,4 @@ extras =
     pyside6: pyside6
 commands_pre =
     pyqt6,pyside6: pip install -U pytest-qt@git+https://github.com/pytest-dev/pytest-qt.git
-commands = pytest --color=yes --cov=superqt --cov-report=xml {posargs}
+commands = pytest --color=yes --cov=superqt --cov-report=xml --pyargs superqt {posargs}
```
tlambert03, line 59: thanks for this.  I'm assuming this is to fix errors like https://github.com/napari/superqt/runs/3203181005

can you help me understand why it fixes that?

Czaki: The `pyargs` told `pytest` that tested part of code is installed python package. 
Earlier problem was connected with two locations of of same test files. They suggest using this in the documentation:
https://docs.pytest.org/en/6.2.x/goodpractices.html#tests-as-part-of-application-code

They also suggest using `src` layout to avoid problems with the identification of which part of the code is tested: 

https://docs.pytest.org/en/6.2.x/goodpractices.html#tests-outside-application-code

Czaki:
> This looks pretty good to me @Czaki! Anything else that needs to be done?

I need to finish the documentation. It`s a little late for me so I will do this tomorrow.

Czaki:
I add a little text to docs and allow to use `Optional[Enum]` as type.

superqt/combobox/_enum_combobox.py
```diff
@@ -0,0 +1,118 @@
...
+        if enum_class is not None:
+            self.setEnumClass(enum_class)
+        self.currentIndexChanged.connect(self._emit_signal)
+
+    def setEnumClass(self, enum: Type[EnumType]):
+        """
+        Set enum class from which members value should be selected
+        """
+        self.clear()
+        if get_origin(enum) is Union:
```
tlambert03, line 48: hmm.  while i think this sort of type annotation introspection makes sense in the context of magicgui (where we're deriving settings from function signatures), it feels a little out of place here.  What would you think about just have another parameter `nullable`, or `allow_none`, or `optional` or something...

```py
def setEnumClass(self, enum: Type[EnumType], nullable=False): ...
```

also, now that I'm looking at typing, I'm not sure the `TypeVar` is working as intended (as opposed to just using `EnumMeta`).  If the goal is to have it know that when you provide a specific `Enum` class, that's what you'll get out of `enumClass()`, or that's what is needed in `setCurrentEnum`, then I'm not sure it works that way (since we're relying on a runtime argument).  Thoughts?

Czaki: > hmm. while i think this sort of type annotation introspection makes sense in the context of magicgui (where we're deriving settings from function signatures), it feels a little out of place here. What would you think about just have another parameter `nullable`, or `allow_none`, or `optional` or something...

changed


> also, now that I'm looking at typing, I'm not sure the `TypeVar` is working as intended (as opposed to just using `EnumMeta`). If the goal is to have it know that when you provide a specific `Enum` class, that's what you'll get out of `enumClass()`, or that's what is needed in `setCurrentEnum`, then I'm not sure it works that way (since we're relying on a runtime argument). Thoughts?

I do not remember why I use `TypeVar` when I implement this in PartSeg. And I use PartSeg implementation as a base for this. It was connected with type checking, but maybe it should be changed to EnumMeta. I will try and see if there will be some warnings.

tlambert03 (approved):
This is looking good to me!  all set @Czaki ?

Czaki:
Hm, maybe need to add information about a new signal in documentation?

tlambert03:
sounds good.  ping me when you're happy with it

tlambert03:
docs look good to me.  Still want more time with this @Czaki?

Czaki:
No. Ready to merge.

tlambert03:
thanks @Czaki!

PR #16, opened 2021-08-05
Add QElidingLabel

### LINKED ISSUE #14: Add Elided Label
The `ElidingLabel` in napari is something I would use in other projects:
https://github.com/napari/napari/blob/master/napari/_qt/widgets/qt_eliding_label.py

tlambert03:
closes #14 

Adds a `QLabel` variant that elides text to fill the available space.  For multi-line Qlabel (`setWordWrap(True)`), it will truncate the number of lines shown, and elide the last line, to fit the available space:

```py
In [1]: from superqt import QElidingLabel

In [2]: text = '''Lorem ipsum dolor sit amet, consectetur adipiscing elit, 
   ...: sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. 
   ...: Ut enim ad minim ven iam, quis nostrud exercitation ullamco laboris
   ...: nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in
   ...: reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.'''

In [3]: wdg = QElidingLabel(text)

In [4]: wdg.resize(260, 80)

In [5]: wdg.show()
```
Untitled

```py
In [6]: wdg.setWordWrap(True)
```

Untitled 2


tlambert03:
@goanpeca, @Czaki, @sofroniewn ... if any of you have a moment to take a quick peek here, i'd appreciate it.  would prefer not to merge anything in this repo without at least one other set of eyes. :)

Czaki:
Maybe set a tooltip with full text?

superqt/_eliding_label.py
```diff
@@ -0,0 +1,105 @@
...
+        """Return `self._text` elided to `width`"""
+        fm = QFontMetrics(self.font())
+        # the 2 is a magic number that prevents the ellipses from going missing
+        # in certain cases (?)
+        width = (width or self.width()) - 2
+        if not self.wordWrap():
+            return fm.elidedText(self._text, self._elide_mode, width)
+
+        # get number of lines we can fit without eliding
+        nlines = self.height() // fm.height() - 1
```
Czaki, line 97: There is no pass of new height (for resize event) so there is a possibility for a wrong number of lines.

sofroniewn (approved):
Didn't try running this, but gave it a quick look over and code looks good

superqt/_eliding_label.py
```diff
@@ -0,0 +1,105 @@
...
+        return self._elide_mode
+
+    def setElideMode(self, mode: Qt.TextElideMode):
+        """Set the elide mode to a Qt.TextElideMode."""
+        self._elide_mode = Qt.TextElideMode(mode)  # type: ignore
+        super().setText(self._elidedText())
+
+    def fullText(self) -> str:
+        """Return the full un-elided text."""
+        return self._text
```
Czaki, line 42: why this instead of plain `text`?

examples/eliding_label.py
```diff
@@ -0,0 +1,12 @@
+from superqt import QElidingLabel
+from superqt.qtcompat.QtWidgets import QApplication
+
+app = QApplication([])
+
+labl = QElidingLabel(
```
goanpeca, line 6: Very minor nitpick, but using `label` feels better than `labl` or `lbl` or `l` :)

superqt/_eliding_label.py
```diff
@@ -0,0 +1,105 @@
...
+    QElidingLabel(parent: Optional[QWidget] = None, f: Qt.WindowFlags = ...)
+    QElidingLabel(text: str, parent: Optional[QWidget] = None, f: Qt.WindowFlags = ...)
+
+    For a multiline eliding label, use `setWordWrap(True)`.  In this case, text
+    will wrap to fit the width, and only the last line will be elided.
+    When `wordWrap()` is True, `sizeHint()` will return the size required to fit the
+    full text.
+    """
+
+    ELIDE_STRING = "…"
```
goanpeca, line 20: Should this be private?

Do we want to be able to configure this? is this locale related?

superqt/_eliding_label.py
goanpeca: Same question 🤔

superqt/_eliding_label.py
```diff
@@ -0,0 +1,105 @@
...
+        """Set the elide mode to a Qt.TextElideMode."""
+        self._elide_mode = Qt.TextElideMode(mode)  # type: ignore
+        super().setText(self._elidedText())
+
+    def fullText(self) -> str:
+        """Return the full un-elided text."""
+        return self._text
+
+    @staticmethod
+    def wrapText(text, width, font=QFont()) -> List[str]:
```
goanpeca, line 45: This feels weird `font=QFont()` 

Why not use a `None` ?

superqt/_eliding_label.py
```diff
@@ -0,0 +1,105 @@
...
+                break
+            ln.setLineWidth(width)
+            start = ln.textStart()
+            lines.append(text[start : start + ln.textLength()])
+        tl.endLayout()
+        return lines
+
+    # Reimplemented QT methods
+
+    def setText(self, txt):
```
goanpeca, line 65: what about text?

goanpeca: txt -> string, text, value

superqt/_eliding_label.py
```diff
@@ -0,0 +1,105 @@
...
+        self._text = txt
+        super().setText(self._elidedText())
+
+    def resizeEvent(self, rEvent: QResizeEvent):
```
goanpeca, line 69: rEvent -> event ?

superqt/_eliding_label.py
```diff
@@ -0,0 +1,105 @@
...
+        super().setText(self._elidedText(rEvent.size().width()))
+        rEvent.accept()
+
+    def setWordWrap(self, wrap: bool) -> None:
```
goanpeca, line 73: Docsrting?

superqt/_eliding_label.py
```diff
@@ -0,0 +1,105 @@
...
+        super().setWordWrap(wrap)
+        super().setText(self._elidedText())
+
+    def sizeHint(self):
```
goanpeca, line 77: Docstring?

goanpeca (changes requested):
Hi @tlambert03 thanks for working on this.

Made some requests and comments.

Overall looks good!

tlambert03:
thank you all!  @goanpeca @Czaki, I agree with the `fullText` to `text()` comments.  The reason I wasn't using text is, as you probably noticed, I'm actually using the public QLabel `text` as the elided text.  (so `text()` would return the elided version).  But I'm convinced by your comments that that's not the right approach, and will refactor to put that logic into the paintEvent rather than the model itself (which is where it belongs anyway)

superqt/_eliding_label.py
tlambert03: can do,  (have to convert to an empty font for the call to QTextLayout, but can do `QTextLayout(text, font or QFont())`)

superqt/_eliding_label.py
tlambert03: in looking into adding these, it turns out that even the `super()` implementation didn't have a docstring, and I don't want to start creating custom docstrings for builtin methods.  currently it gives the same result as the native method which works for me

```py
In [10]: QLabel.sizeHint?
Signature: QLabel.sizeHint(self) -> PySide2.QtCore.QSize
Docstring: sizeHint(self) -> PySide2.QtCore.QSize
Type:      method_descriptor

In [11]: QElidingLabel.sizeHint?
Signature: QElidingLabel.sizeHint(self) -> PySide2.QtCore.QSize
Docstring: sizeHint(self) -> PySide2.QtCore.QSize
```

superqt/_eliding_label.py
tlambert03: according to the [resizeEvent docs](https://doc.qt.io/qt-5/qwidget.html#resizeEvent)

> When resizeEvent() is called, the widget already has its new geometry.

so `self.height()` should already be correct.  really, I didn't need to pass width from the resize event either. perhaps I should remove that?

superqt/_eliding_label.py
tlambert03: added

superqt/_eliding_label.py
tlambert03: great question both :) I thought I wasn't going to be able to do this since I was using the internal model to hold the elided text, but it works out fine.  thanks!

superqt/_eliding_label.py
tlambert03: removed

tlambert03:
I think I addressed everything, and added full test coverage.  Will merge later today unless anyone has more requests. 

will then cut a new release with this and @Czaki's new `QEnumComboBox`.

superqt/_eliding_label.py
goanpeca: 👍🏼

Czaki (approved):

goanpeca (approved):
🚀

PR #21, opened 2021-08-31
Add QMessageHandler context manager

tlambert03:
This adds a convenient way to intercept, store, and re-reroute Qt messages from stdout to a python logger:

```python
# Install a permanent message handler
handler = QMessageHandler()
handler.install()  

# temporarily install
with QMessageHandler() as handler:  
    ...

# re-reoute Qt messages to a python logger at the appropriate level
logger = logging.getLogger(__name__)
with QMessageHandler(logger):  
    ...
```

### DIFF
--- a/superqt/__init__.py
+++ b/superqt/__init__.py
@@ -17,6 +17,7 @@ from .sliders import (
     QRangeSlider,
 )
 from .spinbox import QLargeIntSpinBox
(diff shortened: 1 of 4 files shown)


goanpeca (approved):
Fancy pants! 🚀

PR #27, opened 2021-10-03
[WIP] psygnal backed evented list

alisterburt:
WIP PR for a psygnal backed evented list, working for standard things but child events are not connected so nestable lists will not work

feedback on event naming at this stage would be useful but absolutely no rush

edit: need to go back and fix the diff here


tlambert03:
@alisterburt, gonna close this in favor of https://github.com/tlambert03/psygnal/pull/56

seems like the actual evented containers might go better over there, and then we could use them here

PR #28, opened 2021-10-03
Propagate function name in `ensure_main_thread` and `ensure_object_thread`

Czaki:
This is `superqt` side of the fix bug mentioned in https://github.com/napari/napari/pull/3445. 

Without this fix following code will pass:

```python
class Test(Qobject):
	@ensure_object_thread
	def test_func(self, arg1):
		pass

t = Test()

assert t.test_func.__name__ == "_func"
```

### DIFF
--- a/superqt/utils/_ensure_thread.py
+++ b/superqt/utils/_ensure_thread.py
@@ -63,6 +63,7 @@ def ensure_main_thread(
 
         if hasattr(func, "__name__"):
             _func.__name__ = func.__name__
+        _func.__wrapped__ = func
 
         return _func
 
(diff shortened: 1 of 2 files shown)


Czaki:
I do not understand this coverage report.

jni (approved):

Czaki:
Where release with this will be released? lack of this breaks connection to signal function which consume less arguments than signal propagate (may affect psygnal)

tlambert03:
> Where release with this will be released? 

I can create a new release soon


> lack of this breaks connection to signal function which consume less arguments than signal propagate (may affect psygnal)

how would this affect psygnal?

Czaki:
> how would this affect psygnal? (which doesn't use this package at all)... you mean if someone connected to a function decorated with this?

yes. tomorrow I could try to reproduce it with separated code.

tlambert03:
s'ok... no need.  I believe you, just didn't see the connection at first.  Will release in the next day or two

PR #31, opened 2021-10-13
Threadworker

tlambert03:
the `@thread_worker` decorator is a pretty useful general pattern from napari that stands on its own.   I find myself wanting to use it in non-napari qt projects from time to time.

see https://github.com/napari/napari/pull/1210 for the original Pull request

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -65,4 +65,4 @@ extras =
     pyside6: pyside6
 commands_pre =
     pyqt6,pyside6: pip install -U pytest-qt@git+https://github.com/pytest-dev/pytest-qt.git
-commands = pytest --color=yes --cov=superqt --cov-report=xml {posargs}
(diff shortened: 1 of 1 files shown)


tlambert03:
perfect, will give it a try.  pytest timeout didn't do anything.
it does seem to be the `test_object_thread` test each time.

tlambert03:
thanks!  merged.  let's give it a go!

tlambert03:
two for two!

Czaki:
In my code test hanged on opening QMessageBox (which should not be opened). But I do not see how it could impact in this case. 
I create next PR with dealing delete reference (as it hang on timeout test)

Czaki:
all pass :)

tlambert03:
great!  lemme rerun a couple times.

Czaki:
hangs

PR #32, opened 2021-10-13
move to src layout

tlambert03:
just realized that we're not actually testing a number of things due to the --pyargs superqt flag in tox.ini

i think we might have some failures, we'll see...


Czaki:
> This PR moves us to a src layout and removes `--pyargs superqt` ... it's revealing a few failures. I can fix all the other ones, but can you look at the segfaults we're now getting on pyqt5 in `ensure_thread` (see [example](https://github.com/napari/superqt/runs/3878164756?check_suite_focus=true#step:7:38))

I will try to debug these today.

Issue #30, opened 2021-10-10
Proposal: Collapsible Section Control


MosGeo:
# Justification
A lot of scientific plugins require numerous parameters and settings. It is not user-friendly to show all the settings in the plugin window. I propose a collapsible section control that simplifies hiding unimportant/optional settings in a plugin.

# Possible solutions
1. QTreeWidget and QTreeWidgetItem can be used by inserting a composite widget in QTeeWidgetItem. The main issue with this solution is the indentation which does not look great. Also, it does not have a lot of flexibility.
2. Design a control from scratch using button (to hide and show) and content widget.

# Note
There are possible solutions that are available online (e.g., https://github.com/By0ute/pyqt-collapsible-widget). I think having one in superqt would be better than relying on a separate package.

# Prototype
Here is a simple prototype with nice animation too. it is based on a solution found in StackOverflow for C++ Qt.

![juuUxO3FvN](https://user-images.githubusercontent.com/10503349/136709860-208ec20f-8618-4bd9-9160-bd25c5ee78b7.gif)

```python
'''A collapsible widget'''

import sys
from typing import Union
from qtpy.QtWidgets import QLabel, QLayout
from qtpy.QtCore import QAbstractAnimation, QEasingCurve, \
                        QParallelAnimationGroup, QPropertyAnimation
from qtpy.QtWidgets import QPushButton, QVBoxLayout, QApplication, QWidget
from qtpy.QtCore import Qt

# =================================================================================================
class QCollapsibleSection(QPushButton):
    '''
    A collapsible frame for widgets. This is based on simonxeko solution in 
    https://stackoverflow.com/questions/32476006/how-to-make-an-expandable-collapsable-section-widget-in-qt
    '''
    def __init__(self, content:Union[QWidget,QLayout]=None, min_height=None, **kwargs):
        '''Initializes the component'''
        super().__init__(**kwargs)

        if isinstance(content, QLayout):
            content_widget = QWidget()
            content_widget.setLayout(content)
        else:
            content_widget = content

        self.title = self.text()
        self.setCheckable(True)
        self.setStyleSheet("background:rgba(255, 255, 255, 0); text-align:left;")
        self.animator = QParallelAnimationGroup()
        if content is not None:
            self.set_content_widget(content_widget, min_height)
            self.toggle_hidden()
        self.clicked.connect(self.toggle_hidden)
    # ===========================================
    def toggle_hidden(self) -> None:
        '''Toggle the hidden state of the frame'''
[...]

PR #37, opened 2021-11-03
QCollapsible for Collapsible Section Control

MosGeo:
First time contributor here and a beginner at Qt so bear with me :) This is a pull request to address #30. 

A few notes to consider and discuss
1. The animation is current set by duration which is the standard setting for animations in Qt. Another approach is to use pixel per seconds or something similar.
2.  The default easing curve for the animation is set to InOutCubic ([Easting Curves](https://doc.qt.io/archives/qt-4.8/qeasingcurve.html)).
3. Text alignment is set to left and is set as string. We can use enum instead of string if that is preferred.
4. Icon is currently png. I am not sure what is the best way to handle this item. For example, svg might be better. Also I noticed that there is a font icon pull request so maybe that is better? I don't have experience in this.

Looking forward to your comments!


jni:
I only had a quick skim but this looks very clean to me @MosGeo, thanks! :tada: I think @tlambert03 was moving to using fontawesome for icons rather than bundling .png/.svg icons. @tlambert03 is that implemented? Perhaps this PR is a good test-bed for it?

src/superqt/collapsible/_collapsible.py
```diff
@@ -0,0 +1,161 @@
+"""A collapsible widget to hide and unhide child widgets"""
+import pathlib
+from typing import Union
+
+from PySide2.QtCore import QPropertyAnimation, QVariantAnimation
+from PySide2.QtGui import QTransform
```
tlambert03, line 6: use `superqt.qtcompat` instead of `PySide2` so we can also support PyQt

src/superqt/utils/_animation.py
```diff
@@ -0,0 +1,76 @@
+"""Different animators to be used for animating components"""
+# from math import sin, cos, radians
+
+from ..qtcompat.QtCore import QEasingCurve, QPropertyAnimation, QVariantAnimation
+from ..qtcompat.QtGui import QIcon, QPixmap, QTransform
+from ..qtcompat.QtWidgets import QAbstractButton, QWidget
+
+
+# =================================================================================================
+def create_hide_show_animation(
```
tlambert03, line 10: while I like the idea of these utility functions, I'm not sure they're general enough to live on their own.  That is, this is basically just using the `QPropertyAnimation` api directly, not _adding_ something that Qt doesn't already provide itself.  

Since this function is only used in a single place here, I think for now I'd prefer moving this function into the module where it is used, and prefixing it with an underscore to keep it private (so we're not implying that this is part of the general superqt API just yet)

src/superqt/utils/_animation.py
```diff
@@ -0,0 +1,76 @@
...
+    # Create the animation
+    animation.setStartValue(0)
+    animation.setEasingCurve(easing_curve)
+    animation.setDuration(duration)
+
+    return animation
+
+
+# =================================================================================================
+def create_icon_rotation_animation(
```
tlambert03, line 37: same here, make private in the other module

src/superqt/collapsible/_collapsible.py
```diff
@@ -0,0 +1,161 @@
...
+        self.rotate_animation = create_icon_rotation_animation(
+            self, duration=duration, easing_curve=easing_curve
+        )
+        self.animator.addAnimation(self.hide_show_animation)
+        self.animator.addAnimation(self.rotate_animation)
+
+        # Connect events
+        self.clicked.connect(self._toggleHidden)
+
+    # ===========================================
```
tlambert03, line 103: just personal preference, but can we remove the `# ======` dividers between all methods? and at the end of the file?

tlambert03 (commented):
yeah this looks great @MosGeo !!  very high quality for a first-time contributor new to Qt :)

I agree, it's a bit tricky to know what to do about the png vs svg vs iconfont issue.  I think I'd like to play with this a little bit.  The font icon PR is basically ready to go, but if we used that, we'd need to _also_ include a default font-pack (which that PR doesn't actually do).  It might be a good thing to do ultimately, but for a single caret item, I'm not sure yet. 

Ultimately, I think we need a solution that is easily themeable (at the very least, should be both black and white).  I'll see if we can use QPainter directly to make an object that satisfies all of the needs here without using images or fonts.

more soo!

src/superqt/collapsible/_collapsible.py
MosGeo: ahh... vscode auto-import strikes again :( I have been fighting it but it seems i missed a few.

src/superqt/utils/_animation.py
MosGeo: Agree. I debated this myself. I was not sure and forgot to add it as a note in the PR. migrated.

src/superqt/utils/_animation.py
MosGeo: migrated.

src/superqt/collapsible/_collapsible.py
MosGeo: Nobody actually like these :) I like them while developing since I tend to have lots of commented code so I this keeps me visually organized. I guess it is a remnant from my VB6 days where they have lines between methods.

MosGeo:
@tlambert03 and @jni Thanks for the comments and suggestions. All valid and addressed. Let me know if I need to do anything else.

One thing I will note that right now, there is only soft coupling between QCollapsible and the inner widget. The inner widget needs to be added manually in the main script. It is not a real child of QCollapsible. This means that the location of the inner widget can be anywhere in the application (doesn't have to be at the bottom of the QCollapsible widget.

tlambert03:
thanks again @MosGeo ... I'd like to cut a new release in the next day or two, and it would be nice to include this.  Will you have time to update the tests for the new `__init__` signature?

MosGeo:
Ok. I went on a deep dive with the code and tests. I had a to change a few things:

1. Unfortunately, the right and up arrow symbols in windows do not have the same size. See the picture below. That is why I initially put the up arrow instead of the right arrow :) Right now, using the right and up down arrow together is pretty jarring as it moves the text too. I do not know what is the best solution but with the current situation, I prefer the the up arrow instead of right. what do you think?

image

2. I believe that the default for the control should be collapsed instead of expanded. It is much cleaner interface when you open the software with hidden controls. For changing the default without animation, see item 3.

3. I added three new methods (`quickExpand` and `quickCollapse`, `expanded`). This is basically for collapsing and expanding without animation and checking the status of the control. For example, they can be used for setting up the initial behavior. I don't know if this is the best name for them. Also, I considered modifying the _toggle function with a Boolean to prevent animation but that is an internal function and I wanted to keep it clean. Another option is to have use something like `setChecked` and `isChecked` which is more like Qt but it doesn't convey meaning for the control.

4. There was a logical error. To reproduce, 1) start with an initial state (say collapsed), 2) lock the control, 3) click on the control, 4) unlock the control, 5) click on the control. The expected behavior for this situation is that the control would expand but in the code it was going to collapse again. I added a small if statement to prevent changing the check status for the button if it is locked (in the _toggle function).

5. ` self._toggle_btn.setChecked(forward)`  in `_animate` is superfluous for the most part as the function is only called by clicking the push button which already changes the checked status. But I kept it as it is good to have just incase somebody calls `_animate` from somewhere else.

tlambert03:
Fantastic @MosGeo 

> That is why I initially put the up arrow instead of the right arrow :) Right now, using the right and up down arrow together is pretty jarring as it moves the text too. I do not know what is the best solution but with the current situation, I prefer the the up arrow instead of right. what do you think?

yeah, I agree.  thanks for humoring me, both on trying to animate it, and explaining the width issue.  Let's go back to your original design :)

> I believe that the default for the control should be collapsed instead of expanded

sounds good to me!

> I added three new methods (quickExpand and quickCollapse, expanded). This is basically for collapsing and expanding without animation and checking the status of the control.

I like it!  only question would perhaps be whether the "quick" part should be a parameter to `expand/collapse` instead of a new method?

```py
def expand(self, quick=False):  ...
def collapse(self, quick=False):  ...
def _animate(self, quick=False):  ...
```

thoughts?  If you think `_animate` is then oddly named, that could change too

> There was a logical error.

good catch!

> self._toggle_btn.setChecked(forward) in _animate is superfluous for the most part as the function is only called by clicking the push button which already changes the checked status

yeah, and if someone uses `widget.expand/collapse()` programmatically, couldn't it get out of sync?

MosGeo:
Ok. I think I like the API now with the latest commits (minus the up carret but that can be tackled in another time once a more permanent approach is established for icons in the napari world).

The changes:

- Follow your advise for putting the `quick` parameter as  parameter but with some changes:
-- rename it to `animate` and set the default to `True`.
-- do not include the `animate` parameter in the `_animate` function.
-- Move most of the work of keeping consistency from the `_animate` method to the `collapse` and `expand` methods. Now `animate` is actually just for the animation. It is internal so it does not guarantee consistency when called. I think that is better than duplicating code ('slow down') to guarantee consistency for something that should not be called externally. If you want though, we can guarantee consistency for it too.

- put locks on the externally facing api (`collapse` and `expand`) in addition to `_toggle`.

- Fix for a bug  with test for different screen resolutions. I don't know what is the best approach for this but for now, i am just testing if the height is bigger than zero.

For the icon, I am actually all in for micro-animations in software. They make the controls feel alive. Windows 11 have really nice ones. I just didn't know how to do it so when you mention it, I thought that it would be good time to learn it. So it was not a time wasted or anything. My personal investment in Qt is long term so I want to learn this stuff but I am learning as I go (maybe I should do a udemy course on it or something if I have time). And I didn't mention the width issue initially as I knew that I would have to use icon for animation :) 

Anyway, check out the latest changes. I am now happy with it the way it is and fulfils all my needs for the time being.

src/superqt/collapsible/_collapsible.py
```diff
@@ -0,0 +1,114 @@
...
+        if self._locked:
+            return
+        self._toggle_btn.setChecked(False)
+        self._toggle_btn.setText(
+            self._COLLAPSED + self._toggle_btn.text()[len(self._EXPANDED) :]
+        )
+        if animate:
+            self._animate(QAbstractAnimation.Direction.Backward)
+        else:
+            self._content.setMaximumHeight(0)
```
tlambert03, line 95: the only downside here is that there's now a lot of code duplication, these are essentially the same methods, with a couple very minor things changed... curious what your thoughts are on this versus just renaming `_animate` ... which really was a very internal thing anyway?

tlambert03 (approved):
This works for me!  Couple minor comments, just need to fix the qt6 test, but the other stuff is optional.

Thanks again, love this feature!

src/superqt/collapsible/_collapsible.py
```diff
@@ -0,0 +1,114 @@
...
+    Based on https://stackoverflow.com/a/68141638
+    """
+
+    _EXPANDED = "▼  "
+    _COLLAPSED = "▲  "
+
+    def __init__(
+        self, title: str = "", parent: Optional[QWidget] = None, flags=Qt.WindowFlags()
+    ):
+        super().__init__(parent, flags)
```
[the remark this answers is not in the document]
MosGeo: Looks like the location of the flag has changed in PyQt6. This looks like something that should be handled by qtpy or superqt.qtcompat. If not, and this is needed here we can add this back with conditions to handle the compability issue. For now, I think removing it is the right option 

https://stackoverflow.com/questions/69747328/pyqt6-attributeerror-qt

src/superqt/collapsible/_collapsible.py
MosGeo: @tlambert03 ok. I just merged things into one function to avoid duplication. This means 2 more if statement to execute per expand or collapse but it is probably just a couple of ms extra :) All looks good now

I also removed the top margine from the default content widget. I think it was too far from the title of the collapsible section.

tlambert03: > just a couple of ms extra

probably nanoseconds!
```py
In [1]: def f():
   ...:     return
   ...:

In [2]: def f2():
   ...:     if True:
   ...:         pass
   ...:     if False:
   ...:         pass
   ...:     if True:
   ...:         pass
   ...:     return
   ...:

In [3]: %timeit f()
51.6 ns ± 1.67 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [4]: %timeit f2()
56 ns ± 2.08 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

well worth the code dedup :)

> I also removed the top margine from the default content widget. I think it was too far from the title of the collapsible section.

👍

src/superqt/collapsible/_collapsible.py
```diff
@@ -0,0 +1,107 @@
...
+
+        self._toggle_btn = QPushButton(self._COLLAPSED + title)
+        self._toggle_btn.setCheckable(True)
+        self._toggle_btn.setChecked(False)
+        self._toggle_btn.setStyleSheet("text-align: left; background: transparent;")
+        self._toggle_btn.clicked.connect(self._toggle)
+
+        # frame layout
+        self.setLayout(QVBoxLayout())
+        self.layout().setAlignment(Qt.AlignTop)
```
tlambert03, line 30:
```suggestion
        self.layout().setAlignment(Qt. AlignmentFlag .AlignTop)
```

tests/test_collapsible.py
```diff
@@ -0,0 +1,66 @@
...
+
+    assert wdg1.expanded() is False
+
+
+def test_changing_animation_settings(qtbot):
+    """Quick test for changing animation settings"""
+    wdg = QCollapsible()
+    wdg.setDuration(600)
+    wdg.setEasingCurve(QEasingCurve.InElastic)
+    assert wdg._animation.easingCurve() == QEasingCurve.InElastic
```
tlambert03, line 57:
```suggestion
    wdg.setEasingCurve(QEasingCurve.Type.InElastic)
    assert wdg._animation.easingCurve() == QEasingCurve.Type.InElastic
```

tlambert03:
thanks again @MosGeo!

src/superqt/collapsible/_collapsible.py
MosGeo: Wow didn't know it is this performant.

MosGeo:
Cool. This is exciting. Less code to have in my plugins now once this is released.

Now I gotta decide on my next contribution :)

tlambert03:
> Now I gotta decide on my next contribution :)

please do! :)  it was nice working with you.

PR #39, opened 2021-11-07
Use qtpy, deprecate superqt.qtcompat, drop support for Qt <5.12

tlambert03:
PR that will remove qtcompat and use qtpy again once the qt6 support is realeased.


tlambert03:
this still needs https://github.com/spyder-ide/qtpy/pull/276 and https://github.com/spyder-ide/qtpy/pull/271 before we could use qtpy again

setup.cfg
```diff
@@ -35,6 +35,7 @@ project_urls =
 [options]
 packages = find:
 install_requires =
+    qtpy
```
Czaki, line 38: add minimal version requirement, Without this update of superqt will not force update of qtpy, so compatibility with PySide6/PyQt6 will be missed.

Czaki (changes requested):

setup.cfg
tlambert03: Hmmm, I hesitate to do that in case someone wanted to use an older version of qtpy with qt 5.15?

Its not that superqt required 2.0+, it's that PyQt6 does... thoughts on that?  So it could make some environment resolutions more difficult if we do that.

Thoughts?

Czaki: is there release from the time of introduction qtcompat? If yes hen someone may have code that uses `superqt`,   and have an older version of `qtpy` in the environment, then after the update, the code will be broken. But maybe this is enough small that put information in readme will be enough.

PR #40, opened 2021-11-09
Add qt_set_trace and install_qt_breakpoint

tlambert03:
This PR adds a few conveniences around using a debugger with Qt.

to add a breakpoint in Qt code, use it like `pdb.set_trace()`

```py
from superqt.utils import qt_set_trace; qt_set_trace()
```

This will stop the event loop so you don't get an endless chain of 
```
QCoreApplication::exec: The event loop is already running
QCoreApplication::exec: The event loop is already running
QCoreApplication::exec: The event loop is already running
```

or to permanently install it so that you can use the builtin `breakpoint()` anywhere

```py
from superqt.utils import install_qt_breakpoint

install_qt_breakpoint()
```

### DIFF
--- a/src/superqt/utils/__init__.py
+++ b/src/superqt/utils/__init__.py
@@ -8,8 +8,12 @@ __all__ = (
     "QMessageHandler",
     "thread_worker",
     "WorkerBase",
+    "install_qt_breakpoint",
+    "uninstall_qt_breakpoint",
+    "qt_set_trace",
 )
 
+from ._breakpoint import qt_set_trace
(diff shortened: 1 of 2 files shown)


tlambert03:
holding on this for now.  
first: it's only necessary for PyQt5 (pyside already works fine)... second: i want to add a way to accept an arbitrary debugger's `set_trace` function.  The main issue I'm running into (if I don't use the subclass approach here) is that the breakpoint starts at the wrong frame.  I essentially want to remove the `QtCore.pyqtRemoveInputHook()` from the frame stack before starting the debugger...

Issue #45, opened 2021-11-23
Older version typing-extensions causing trouble with superqt


AhmetCanSolak:
**Describe the bug**
Older version of typing-extensions causing trouble with superqt during import time. I am happy to contribute a fix upon a decision together about how to address this(pinning typing-extensions version or something else?).

**To Reproduce**
Install tensorflow==2.6.0(this installs typing-extensions==3.7.4.3), install superqt(or napari), and import superqt(or napari).

**Expected behavior**
superqt doesn't cause trouble with imports in an ideal world. With latest typing-extensions(or tensorflow==2.7.0) there is no problem.

**Screenshots**

```
superqt/utils/_qthreading.py", line 23, in <module>
    from typing_extensions import Literal, ParamSpec
ImportError: cannot import name 'ParamSpec' from 'typing_extensions' 
```

**Desktop (please complete the following information):**
 - OS with version [e.g macOS 10.15.7] OSX Mojave, Linux 18.04
 - Qt Backend [e.g PyQt5, PySide2] PYQT5
 - Python version Python3.9

tlambert03:
ahh, thanks @AhmetCanSolak ... I think we need `typing-extensions>=3.10.0.0`

tlambert03:
bummer... looks like tensorflow pinned to `~=3.7.4` (meaning we won't be able to use anything >3.7) ... that's rather annoying for something like typing-extensions, which is just a backport.  there's very little reason to pin!  ... I can refactor to not use ParamSpec, but we loose some nice typing :/  argh

tlambert03:
does it work if you just manually `pip install typing-extensions -U`?

tlambert03:
> I think we need typing-extensions>=3.10.0.0

maybe just make a PR to this effect... I should have done that if I was using ParamSpec anyway

AhmetCanSolak:
yes they pinned :( honestly I like ParamSpec stuff. I prefer we add `typing-extensions>=3.10.0.0`

`pip install typing-extensions -U` gives dependency resolver conflict as tensorflow pinned to 3.7.4

tlambert03:
> gives dependency resolver conflict as tensorflow pinned to 3.7.4

but it installs right?  or are you on conda?

AhmetCanSolak:
> > gives dependency resolver conflict as tensorflow pinned to 3.7.4
> 
> but it installs right, just with red text? ~or are you on conda?~ (no you said pip)

Yes it installs but gives red warning text.

[...]

PR #46, opened 2021-11-23
typing-extensions version pinning

AhmetCanSolak:
As we discussed in #45 , I am making a follow up PR here.

- initial commit


tlambert03:
Thanks, will release a patch soon

Issue #48, opened 2021-12-23
Bug: signal type overrides broken on PySide6.2.1.1


tlambert03:
**Describe the bug**
haven't figured out what yet... but it's all of the failures in [this test](https://github.com/napari/superqt/runs/4585637222?check_suite_focus=true)

```pytb
________________________________________________________________________________
Traceback (most recent call last):
  File "/Users/talley/Dropbox (HMS)/Python/forks/superqt/src/superqt/sliders/_labeled.py", line 313, in _on_value_changed
    if len(v) != len(self._handle_labels):
TypeError: object of type 'int' has no len()
```

that's in the RangeSlider, where value should be a tuple... so i wonder whether class inheritance has somehow been slightly altered

tlambert03:
ok, looks like PySide6.2 did something fundamental with signal types.  It used to be possible to override the _type_ of a signal in a subclass.  for example, this worked in PySide2 and Pyside6<6.2:

```py
class MyWidget(QSlider):
    valueChanged = Signal(tuple)

# valueChanged would be a tuple
```

in 6.2 however, it looks like that's no longer possible.  if you look at the value emitted in 6.2, even if you do `mywidget.valueChanged.emit((20, 30))`, the actual emitted value will be `0` :/

I created an issue [here](https://bugreports.qt.io/browse/PYSIDE-1751) ... but I suspect this _may_ be a "will not fix", meaning we might be stuck with some rather unfortunate breaking API changes for the range slider

eyalk11:
Got same bug on 6.2.1.1 . 6.2.0 works.

Czaki:
what happens if you specify `Signal(object)`?

I use `object` workaround for Enum in PartSeg 

https://github.com/4DNucleome/PartSeg/blob/4451f4ce2a5e2861d90a69c0977354b19da6b2ce/package/PartSeg/common_gui/universal_gui_part.py#L31

https://github.com/4DNucleome/PartSeg/blob/4451f4ce2a5e2861d90a69c0977354b19da6b2ce/package/PartSeg/common_gui/universal_gui_part.py#L68

At this moment I do not remember error, but it was something similar.

tlambert03:
closed by #51

PR #51, opened 2022-01-02
Ugly but functional workaround for pyside6.2.1 breakages

tlambert03:
Trying this out on CI.  This gets around the problem described in #48 by overriding signal instance attributes at instantiation time.  It's really fugly, but it's the only way I can think of at the moment to deal with the changes in pyside6.2  without breaking superqt's API (without deprecation).  We can discuss whether we should generally change the signals API, but I think that should go elsewhere:

This uses the following pattern, which appears to work:

```py
class NewSlider(QSlider):
    _fvalueChanged = Signal(float)
    def __init__(self):
        super().__init__()
        self.valueChanged = self._fvalueChanged  # need to do the override after super().__init__
```

### DIFF
--- a/src/superqt/sliders/_generic_range_slider.py
+++ b/src/superqt/sliders/_generic_range_slider.py
@@ -32,16 +32,17 @@ class _GenericRangeSlider(_GenericSlider[Tuple], Generic[_T]):
     """
 
     # Emitted when the slider value has changed, with the new slider values
-    valueChanged = Signal(tuple)
+    _valuesChanged = Signal(tuple)
 
     # Emitted when sliderDown is true and the slider moves
     # This usually happens when the user is dragging the slider
     # The value is the positions of *all* handles.
-    sliderMoved = Signal(tuple)
+    _slidersMoved = Signal(tuple)
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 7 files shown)


tlambert03:
@goanpeca, @Czaki, curious to get your thoughts on this PR.

I'm aware it's a ghastly looking pattern, but it's the only thing I've been able to come up with so far that doesn't cause a breaking API change for superqt uses with pyside6.2.1.1  (see https://github.com/napari/superqt/issues/48#issuecomment-1000445539 for details)

alternatively... we could release a `!=6.2.1.1` dependency patch, and see how [this issue at PySide](https://bugreports.qt.io/browse/PYSIDE-1751) resolves?

goanpeca:
>I'm aware it's a ghastly looking pattern, but it's the only thing I've been able to come up with so far that doesn't cause a breaking API change for superqt users with pyside6.2.1.1 (see #48 (comment) for details)

I think it is a pretty decent solution in the meantime @tlambert03, :)

Go for it!

goanpeca (approved):
I actually like the solution @tlambert03 , go for it!

Czaki:
I do not see any better solution which does not need reimplement classes to not Inherit QSlider or QAbstactSlider.

PR #69, opened 2022-02-17
Add `signals_blocked` util

tlambert03:
context manager that blocks signal emission from an object

```python
from superqt.utils import signals_blocked

with signals_blocked(obj):
    obj.signal.emit()  # will be blocked
```


ianhi (approved):
nice!

PR #79, opened 2022-04-16
Fix CSS for Collapsible

MinmoTech:
The button used for the Collapsible previously showed a black on
archlinux.
This fixes it to display properly.
Collapsilbe before fix:
image

Collapsilbe after fix:
image

Closes https://github.com/napari/superqt/issues/78


MinmoTech:
Could someone please test this on windows (and macos) before merging? :)

tlambert03 (approved):
works for me!  thanks for the fix!

PR #88, opened 2022-05-10
Add code syntax highlight utils

Czaki:
This PR adds code syntax highlight first introduced in https://github.com/napari/napari/pull/4414
Zrzut ekranu z 2022-05-10 19-13-19

### DIFF
--- a/docs/utils.md
+++ b/docs/utils.md
@@ -8,3 +8,11 @@ that can be used to highlight code in a QTextEdit.
 Code lexer and available styles are from [`pygments`](https://pygments.org/) python library
 List of available languages are available [here](https://pygments.org/languages/).
 List of available styles are available [here](https://pygments.org/styles/).
+
+If you would like to use this in napari widget,
+then you may tant to use syntax style from napari settings.
+To get current syntax style, use
+`syntax_style = get_theme(get_settings().appearance.theme, as_dict=False).syntax_style`
+
+The function `get_theme` is defined in `napari.utils.theme` module.
+The function `get_settings` is defined in `napari.settings` module.
--- a/docs/utils.md
+++ b/docs/utils.md
@@ -8,11 +8,3 @@ that can be used to highlight code in a QTextEdit.
 Code lexer and available styles are from [`pygments`](https://pygments.org/) python library
 List of available languages are available [here](https://pygments.org/languages/).
 List of available styles are available [here](https://pygments.org/styles/).
-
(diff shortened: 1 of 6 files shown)


docs/utils.md
```diff
@@ -0,0 +1,18 @@
...
+List of available languages are available [here](https://pygments.org/languages/).
+List of available styles are available [here](https://pygments.org/styles/).
+
+If you would like to use this in napari widget,
+then you may tant to use syntax style from napari settings.
+To get current syntax style, use
+`syntax_style = get_theme(get_settings().appearance.theme, as_dict=False).syntax_style`
+
+The function `get_theme` is defined in `napari.utils.theme` module.
+The function `get_settings` is defined in `napari.settings` module.
```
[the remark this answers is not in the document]
tlambert03: I'm not sure I'd include this doc here.  It would be pretty easy for this to go out of sync with the napari API, and we don't otherwise mention napari in this repo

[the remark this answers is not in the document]
Czaki: So open separate PR to napari with a description of how to handle change theme in `napari`?

examples/code_highlight.py
```diff
@@ -0,0 +1,32 @@
...
+from superqt.utils import CodeSyntaxHighlight
+
+app = QApplication([])
+
+text_area = QTextEdit()
+
+highlight = CodeSyntaxHighlight(text_area.document(), "python", "monokai")
+
+palette = text_area.palette()
+palette.setColor(QPalette.Base, QColor(highlight.background_color))
```
tlambert03, line 13: this seems like it might be something that happens a lot huh?  Do you think we should add a `set_background` parameter to `CodeSyntaxHighlight` that does this automatically?  (default could be either True or False)

Czaki: You mean to pass `QTextEdit` instead of `QTextDocument` to `CodeSyntaxHighlight` constructor? 
Or allow both types and use `set_background` only if `QTextEdit` is passed?

Czaki: Or maybe add method `setTextEditBackground(self, test_edit: QTextEdit)` ?

tlambert03: ohhh, never mind me :)  I missed the fact that it's `text_area.document()`
let's leave that bit of magic for later.  The example is good enough for now

tlambert03 (approved):

tlambert03:
thanks!

PR #106, opened 2022-08-03
Fix float value error in pyqt configuration

mstabrin:
In the pyqt5 configuration, the `QLabeledSlider` throws an error due to `QDoubleSpinBox` returning a float rather than an int.

Here I propose to add a proxy `_setValue` function which does the conversion :)

### DIFF
--- a/src/superqt/sliders/_labeled.py
+++ b/src/superqt/sliders/_labeled.py
@@ -128,7 +128,7 @@ class QLabeledSlider(_SliderProxy, QAbstractSlider):
         super().__init__(parent)
 
         self._slider = self._slider_class()
-        self._label = SliderLabel(self._slider, connect=self._slider.setValue)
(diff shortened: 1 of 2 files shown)


tlambert03 (approved):
looks great.  thanks for your first contribution here @mstabrin!

mstabrin:
Hello @tlambert03, I was wondering if there is a fixed release cycle for superqt or if releases are created on demand.
I am just wondering when I can expect this change to be available via pip to fix the current problems we face within napari with the Point size slider :)

tlambert03:
they're done as needed. sorry for the delay on this one.  will release today

tlambert03:
oh actually... looks like I tried to release 24 days ago and missed that one of the jobs failed: https://github.com/napari/superqt/actions/runs/2727679828 

rerunning now

PR #112, opened 2022-08-26
Bump qtpy to at least 1.10

Czaki:
To import `SignalInstance` from `qtpy.QtCore` the minimum version of qtpy is `1.10`

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -37,7 +37,7 @@ packages = find:
 install_requires =
     packaging
     pygments>=2.4.0
-    qtpy>=1.1.0
+    qtpy>=1.10.0
     typing-extensions
 python_requires = >=3.7
 include_package_data = True


tlambert03:
as far as I can tell, `SignalInstance` is only imported in one place, and under a TYPE_CHECKING clause: 

https://github.com/napari/superqt/blob/a3bd0d0edf046680bde3dabd653138749a02a20d/src/superqt/utils/_throttler.py#L38

so it shouldn't be required for runtime... what did you run into?

Czaki:
Ahh. The old release does not have a guard check, 

https://github.com/napari/napari/runs/8043776391?check_suite_focus=true#step:8:80

tlambert03:
ah, yep.  that was fixed in https://github.com/napari/superqt/pull/70/

tlambert03:
you could bump napari's requirements to https://github.com/napari/superqt/releases/tag/v0.3.1 if you want

Czaki:
> you could bump napari's requirements to [`v0.3.1` (release)](https://github.com/napari/superqt/releases/tag/v0.3.1) if you want

I just bump `qtpy`, so it solves the problem from this side.

Issue #118, opened 2022-10-02
Enum Combo tests failing on PyQt6


tlambert03:
@Czaki ... if/when you have a moment, some of the `test_enum_comb_box.py` tests appear to be failing recently, only on Qt6 ... just a small change in the expected text:

https://github.com/napari/superqt/actions/runs/3169507004/jobs/5163067458#step:7:466

PR #119, opened 2022-10-03
Simple workaround for pyside 6

Czaki:
I found that recent version of pyside6 overwrite __str__ method of `Enum` class, so checking if str method belongs either to `enum` or `shibokensupport` modules solves this problem. 

But I'm not sure if it is a goo d idea as it may impact on some related enums. 

Maybe the best option will be to wait on the next PySide6 release? Maybe they will solve it?

fixes: #118


Czaki:
imageio just released a new version 
https://pypi.org/project/imageio/#history

https://github.com/imageio/imageio/issues/887

tlambert03:
this works fine for me.  we can iterate if/when pyside changes in the future.  thanks @Czaki!

tlambert03 (approved):

PR #128, opened 2022-10-10
fix: allow submodule imports

kne42:
fix a bug where submodules weren't able to be imported because the top-level module `__getattr__` was raising an `ImportError` instead of an `AttributeError`

raising an `AttributeError` allows the built-in importer to catch it and then check for a submodule that can be imported before raising an `ImportError` itself if nothing is found

see [this failing job](https://github.com/napari/app-model/actions/runs/3175202057/jobs/5268843725#step:6:76)


tlambert03:
thanks @kne42  ... I'm actually working on a fix right now that would preserve the `from superqt import fonticon` behavior

tlambert03:
oh! nevermind :joy: I'm overcomplicating it... thanks for this!

tlambert03 (approved):

kne42:
What's the desired behaviour? This is only happening because we raised the wrong type of error, e.g. it will still raise an `ImportError` at the end of the day in the correct cases:
```python
In [1]: from superqt import fonticon

In [2]: from superqt import asdf
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
Input In [2], in <cell line: 1>()
----> 1 from superqt import asdf

ImportError: cannot import name 'asdf' from 'superqt' (/Users/kiraevans/Projects/superqt/src/superqt/__init__.py)
```

kne42:
oh yeah no worries, overcomplicating things is my main vice so i understand 😅

tlambert03:
yanked 0.3.7,  releasing 0.3.8 now and rerunning app-model tests

PR #142, opened 2022-11-26
Add signal to QCollapsible

ppwadhwa:
This PR adds a signal that is emitted when the toggle button to expand/collapse the widget is pressed.  This signal was needed in https://github.com/napari/napari/pull/5198 in order to resize another widget when the collapsible widget expands/collapses.

### DIFF
--- a/src/superqt/collapsible/_collapsible.py
+++ b/src/superqt/collapsible/_collapsible.py
@@ -1,7 +1,15 @@
 """A collapsible widget to hide and unhide child widgets"""
 from typing import Optional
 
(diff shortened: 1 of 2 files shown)


tlambert03 (commented):
Seems reasonable!

However, It looks to me like this signal will only emit if someone clicks the button, but not on a programmatic change using expand or collapse.

Can you check/test/fix that?

tlambert03 (approved):
thanks @ppwadhwa.

I moved the signal to the `_expand_collapse` (since both collapse and expand just call that method, it's the better place to put the signal, and also accounts for whether the collapsible is locked)

ppwadhwa:
thanks, @tlambert03 !!

PR #144, opened 2022-11-30
Move QCollapsible toggle signal emit

ppwadhwa:
This PR will only emit the toggle signal when the animate flag is True.  I realized that the signal was being emitted 3 times when it was not inside the `animate` if statement.  I don't know if its needed when `animate = False`, but for my purposes, I would prefer if this signal is not emitted several times per mouse click.

### DIFF
--- a/src/superqt/collapsible/_collapsible.py
+++ b/src/superqt/collapsible/_collapsible.py
@@ -132,10 +132,9 @@ class QCollapsible(QFrame):
             self._animation.setEndValue(_content_height)
             self._is_animating = True
             self._animation.start()
+            self.toggled.emit(direction == QPropertyAnimation.Direction.Forward)
         else:
             self._content.setMaximumHeight(_content_height if forward else 0)
-
-        self.toggled.emit(direction == QPropertyAnimation.Direction.Forward)
 
     def _toggle(self):
         self.expand() if self.isExpanded() else self.collapse()
--- a/src/superqt/collapsible/_collapsible.py
+++ b/src/superqt/collapsible/_collapsible.py
@@ -120,6 +120,11 @@ class QCollapsible(QFrame):
         animate: bool = True,
         emit: bool = True,
     ):
+        """Set values for the widget based on whether it is expanding or collapsing.
+
+        An emit flag is included so that the toggle signal is only called once (it
+        was being emitted a few times via eventFilter when the widget was expanding
+        previously)."""
         if self._locked:
             return
 
--- a/src/superqt/collapsible/_collapsible.py
+++ b/src/superqt/collapsible/_collapsible.py
@@ -115,7 +115,10 @@ class QCollapsible(QFrame):
         return self._locked
 
     def _expand_collapse(
-        self, direction: QPropertyAnimation.Direction, animate: bool = True
+        self,
+        direction: QPropertyAnimation.Direction,
+        animate: bool = True,
+        emit: bool = True,
     ):
         if self._locked:
             return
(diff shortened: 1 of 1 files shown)


ppwadhwa:
@tlambert03 I am not able to tag you for review, so this is in place of that.  :)

thanks for looking!

src/superqt/collapsible/_collapsible.py
```diff
@@ -132,11 +132,10 @@ def _expand_collapse(
             self._animation.setEndValue(_content_height)
             self._is_animating = True
             self._animation.start()
+            self.toggled.emit(direction == QPropertyAnimation.Direction.Forward)
         else:
             self._content.setMaximumHeight(_content_height if forward else 0)
 
-        self.toggled.emit(direction == QPropertyAnimation.Direction.Forward)
```
goanpeca, line 138: But if animate is  `False` then it would not be emitted at all.

Do you know why it was being called three times before @ppwadhwa or @tlambert03 ?

goanpeca: Ok what if have something like this, that way we never emit signals when calling expand_collapse from within the eventFilter
```python
...

    def _expand_collapse(
        self, direction: QPropertyAnimation.Direction, animate: bool = True, emit: bool = True
    ):
        if self._locked:
            return

        forward = direction == QPropertyAnimation.Direction.Forward
        text = self._EXPANDED if forward else self._COLLAPSED

        self._toggle_btn.setChecked(forward)
        self._toggle_btn.setText(text + self._toggle_btn.text()[len(self._EXPANDED) :])

        _content_height = self._content.sizeHint().height() + 10
        if animate:
            self._animation.setDirection(direction)
            self._animation.setEndValue(_content_height)
            self._is_animating = True
            self._animation.start()
        else:
            self._content.setMaximumHeight(_content_height if forward else 0)

        if emit:
            self.toggled.emit(direction == QPropertyAnimation.Direction.Forward)

    def _toggle(self):
        self.expand() if self.isExpanded() else self.collapse()

    def eventFilter(self, a0: QObject, a1: QEvent) -> bool:
        """If a child widget resizes, we need to update our expanded height."""
        if (
            a1.type() == QEvent.Type.Resize
            and self.isExpanded()
            and not self._is_animating
        ):
            # Since this can be called several times when child widgets resize, we need to
            # ensure we do not emit the signal while we are animating
            self._expand_collapse(QPropertyAnimation.Direction.Forward, animate=False, emit=False)
        return False

...

```

goanpeca:
Ok what if have something like this, that way we never emit signals when calling expand_collapse from within the eventFilter
```python
    def _expand_collapse(
        self, direction: QPropertyAnimation.Direction, animate: bool = True, emit: bool = True
    ):
        if self._locked:
            return

        forward = direction == QPropertyAnimation.Direction.Forward
        text = self._EXPANDED if forward else self._COLLAPSED

        self._toggle_btn.setChecked(forward)
        self._toggle_btn.setText(text + self._toggle_btn.text()[len(self._EXPANDED) :])

        _content_height = self._content.sizeHint().height() + 10
        if animate:
            self._animation.setDirection(direction)
            self._animation.setEndValue(_content_height)
            self._is_animating = True
            self._animation.start()
        else:
            self._content.setMaximumHeight(_content_height if forward else 0)

        if emit:
            self.toggled.emit(direction == QPropertyAnimation.Direction.Forward)

    def _toggle(self):
        self.expand() if self.isExpanded() else self.collapse()

    def eventFilter(self, a0: QObject, a1: QEvent) -> bool:
        """If a child widget resizes, we need to update our expanded height."""
        if (
            a1.type() == QEvent.Type.Resize
            and self.isExpanded()
            and not self._is_animating
        ):
            # Since this can be called several times when child widgets resize, we need to
            # ensure we do not emit the signal while we are animating
            self._expand_collapse(QPropertyAnimation.Direction.Forward, animate=False, emit=False)
        return False
```

I would also add a docstring to the `_expand_collapse` to explain things a vit for future reference :) @ppwadhwa

goanpeca (approved):
Thanks for the quick fix @ppwadhwa 🚀

PR #154, opened 2023-03-29
Add `QElidingLineEdit` class for elidable `QLineEdit`s

### LINKED ISSUE #152: Add elided `QLineEdit` class
Hi, while working on https://github.com/napari/napari/pull/5640 I ended up checking a way to add eliding capabilities to a `QLineEdit` at https://github.com/napari/napari/pull/5665 

Following the discussion (https://github.com/napari/napari/pull/5665#issuecomment-1483374558), could it be worthy to add here something like an [`ElidingLineEdit` class](https://github.com/napari/napari/blob/d9e3b5c604116cfcb8eeed2527ef803828172806/napari/_qt/widgets/qt_dims_slider.py#L31-L107)?

dalthviz:
Example code:

```python
from superqt import QElidingLineEdit

text = "Lorem ipsum dolor sit amet, consectetur adipiscing elit"
wdg = QElidingLineEdit(text)
wdg.resize(100, 20)
wdg.show()
```

Preview:

imagen

eliding_line_edit

closes #152


tlambert03:
thank you @dalthviz!

the behavior and tests look great.  I'm a bit unhappy with the inheritance pattern at the moment, but I can see that it's nothing you've done wrong (more of an annoyance about the fact the "text-having" parts of QLabel and QLineEdit don't share any inheritance...)  It's the fact that we need to have the same methods in the subclasses:

```python
    # appears more or less the same in both subclasses
    def setElideMode(self, mode: Qt.TextElideMode) -> None:
        """Set the elide mode to a Qt.TextElideMode."""
        super().setElideMode(mode)
        super().setText(self._elidedText())
```

i'd like to play with it briefly to see whether there's anything else I can think of.  If not, I might just consider not using the same base class

(don't worry about the pyside6 and napari tests at the moment either)

tlambert03:
hey @dalthviz, very sorry about the delay here.  Shall we just get this in?
While it's nice to have a clean internal pattern, all I really care about is the public API and tests.  we can always iterate on the implementation.  So, are you happy with the state of things from that perspective?  Is it as you would want to use it over in (e.g.) napari?

dalthviz:
Hi @tlambert03 ! No problem! And I think the functionality provided works for what I wanted to achieve over napari indeed, so merging this could be nice :+1:

Thank you for checking this again and for sure open to improve in a follow up PR the implementation details as needed!

PR #157, opened 2023-04-06
fix: fix pyside6 by removing typing

tlambert03:
frustratingly, it looks like I can no longer inherit from both `QSlider` and `typing.Generic` with pyside6 6.5

haven't had time to look into it farther, but this would fix stuff on pyside6


tlambert03:
still not consistently fixing...  sometimes all tests pass, but then segfault or something on cleanup

tlambert03:
yeah, playing with that in https://github.com/pyapp-kit/superqt/pull/157

tlambert03:
lol ... sorry, I thought we were still in the other PR :joy:  thanks

PR #158, opened 2023-04-15
Searchable tree widget from a mapping

### LINKED ISSUE #12: Add TreeViz widget to allow for expandable tree like metadata structures
Create a widget that when given a TreeLike structure (`Dict`, `xml.etree.ElementTree`, anything else?) display a collapsed view of the structure and allow click on the branches to expand.

Much like existing `ome-types` tree viewer but generalized to all `TreeLike` structures.
image

Didn't get around to it last week but going to try to dedicate some time to it this week. Sorry about that. If I could self assign the issue, I would :joy: 

---

Original discussion from zulip:

@JacksonMaxfield
> @**Talley Lambert** I am tempted to port your `OMETree` widget to some other repo as a more generalized "given a tree like structure (etree.ElementTree or Dict), add widget component" so that we can just toss CZI and LIF metadata in there too. Any thoughts as to where this should go? Is this a stupid idea / someone else has already done that?
>
> This: https://github.com/tlambert03/ome-types/blob/9954ae63c162ce91d516167009590bcc20a77174/src/ome_types/widgets.py#L21

@tlambert03 
> yeah, I agree, it's a very generic component... how about our new `superqt` repo? https://github.com/napari/superqt
> the goal there is to collect things that have no dependencies but "some qt backend" in the environment... so it seems well suited.  We'll just want to modify some of the method names and stuff to try to make it feel as native as possible
>
> is that something you'd want to help out with?

@JacksonMaxfield
> Yea that's what I was thinking as well. Figured superqt would be the way to go.
> 
> Happy to help on that. I _may_ have some time tonight even!

andy-sweet:
This adds a searchable tree widget that can be initialized from a mapping. Mappings and iterables (other than strings) are traversed and populated in the tree. The tree has two columns.

The first column is either the key of a mapping or the enumerated index of an iterable, which can be searched based on regular expression. An item is shown if it, any of its ancestors, or any of its descendants match the expression. Ancestors are needed because tree items are only shown if all their ancestors are. Descendants are needed because if the user if look for a key that 

The second column is either the `str()` value (for terminal types) or the type itself (e.g. `dict`) for non-terminal iterables.

I think this closes #12 , though I have not tried to reimplement [OMETree](https://github.com/tlambert03/ome-types/blob/9954ae63c162ce91d516167009590bcc20a77174/src/ome_types/widgets.py#L21) using this yet.

Here's an example of using the tree on [an OME-NGFF `multiscales` group `.zattrs` JSON file from Zenodo](https://zenodo.org/record/7120354#.ZD3OjezML0o), which is easy to setup with just a `json.load`.

https://user-images.githubusercontent.com/2608297/232627769-9c74592d-8a4c-4d9f-af0c-e545e75e87b5.mp4

For XML, one would have to convert the element tree into a dictionary first (e.g. using [`xmltodict`].(https://github.com/martinblech/xmltodict)).

### DIFF
--- a/src/superqt/selection/_searchable_tree_widget.py
+++ b/src/superqt/selection/_searchable_tree_widget.py
@@ -20,7 +20,7 @@ class QSearchableTreeWidget(QWidget):
         layout.addWidget(self.tree_widget)
         self.setLayout(layout)
 
-    def _setData(self, data: dict) -> None:
+    def setData(self, data: dict) -> None:
         self.tree_widget.clear()
         top_level_items = [_to_item(k, v) for k, v in data.items()]
         self.tree_widget.addTopLevelItems(top_level_items)
@@ -34,7 +34,7 @@ class QSearchableTreeWidget(QWidget):
     @classmethod
     def fromDict(cls, data: dict, *, parent: QWidget = None) -> 'QSearchableTreeWidget':
         widget = cls(parent)
-        widget._setData(data)
+        widget.setData(data)
         return widget
 
 
--- a/src/superqt/selection/_searchable_tree_widget.py
+++ b/src/superqt/selection/_searchable_tree_widget.py
@@ -13,14 +13,14 @@ class QSearchableTreeWidget(QWidget):
     If the mapping changes, the easiest way to update this is by calling `setData`.
 
     The tree can be searched by entering a regular expression pattern
-    into the `filter_widget` line edit. An item is only shown if its key
+    into the `filter` line edit. An item is only shown if its key
     or any of its ancestors' or descendants' keys match this pattern.
 
     Attributes
     ----------
-    tree_widget : QTreeWidget
+    tree : QTreeWidget
         Shows the mapping as a tree of items.
-    filter_widget : QLineEdit
+    filter : QLineEdit
         Used to filter items in the tree by matching their key against a
         regular expression.
     """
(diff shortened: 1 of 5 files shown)


tlambert03:
this is great.  thanks @andy-sweet!  I think the code is plenty readable and appropriately general for a first pass.  If you can just add some tests I'm happy to get this is roughly as is

andy-sweet:
I decided to make this a bit more general, refined the filtering behavior, and added some docs. I'll likely add some tests early next week and may do a little more polishing here. Will mark this ready for review when that's done.

tlambert03:
lemme know when ready... don't worry about the qt-specific test fails.  trying to deal with that in another PR

andy-sweet:
> lemme know when ready... don't worry about the backend-specific test fails. trying to deal with that in another PR

Ready now. I changed a few things since you looked at the draft last.

Probably the biggest change is that I don't evaluate `str()` for mappings or iterable values and instead just show the type (e.g. `dict`). Without this change, large trees tend to be hard to read visually (i.e. lots of similar dictionary strings in the second column) and there are obvious potential performance concerns.

I also swapped the example for something a little more concise and JSON-like. I also wrote examples for JSON and XML, but didn't include them yet to avoid adding too much. Let me know if you'd like them.

andy-sweet:
I also started wondering if the search term should affect the visibility of the items, or if we use it scroll to the first/best matching item. In all cases, there is a question about what to do with matches that are current collapsed and therefore not visible to the user. Expanding is straightforward, but reverting to expansion/collapsed states after that is difficult.

goanpeca:
>I also started wondering if the search term should affect the visibility of the items, or if we use it scroll to the first/best matching item. In all cases, there is a question about what to do with matches that are current collapsed and therefore not visible to the user. Expanding is straightforward, but reverting to expansion/collapsed states after that is difficult

One could even get fancy and do fuzzy matching and highlight the found items 😅

andy-sweet:
Another thought I had here is that this PR is contributing two things.

1\. Searching/filtering a tree widget by the first column.
2\. Creating and (wholesale) updating a tree widget from a mapping.

It could be worth splitting those two contributions, as there may be some desire for creating/updating a tree widget from a nested list too.

src/superqt/selection/_searchable_tree_widget.py
```diff
@@ -0,0 +1,109 @@
...
+    item: QTreeWidgetItem, expression: QRegularExpression, ancestor_match: bool = False
+) -> bool:
+    """Recursively update the visibility of a tree item based on an expression.
+
+    An item is visible if it, any of its ancestors, or any of its descendants
+    match the expression.
+    The text of an item's first column is used to match the expression.
+    Returns True if the item is visible, False otherwise.
+    """
+    text = item.text(0)
```
tlambert03, line 100: what do you think about searching within the values as well?  might be slightly surprising to search for a substring that you know appears in one of the values and not see it appear?

src/superqt/selection/_searchable_tree_widget.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+
+    def __init__(self, parent=None):
+        super().__init__(parent)
+
+        self.tree: QTreeWidget = QTreeWidget(self)
+        self.tree.setHeaderLabels(("Key", "Value"))
+        self.tree.setUniformRowHeights(True)
+
+        self.filter: QLineEdit = QLineEdit(self)
```
tlambert03, line 35:
```suggestion
        self.filter: QLineEdit = QLineEdit(self)
        self.filter.setClearButtonEnabled(True)
```

tlambert03 (approved):
looks great.  only two minor comments, then happy to merge when you're ready

tlambert03:
> It could be worth splitting those two contributions, as there may be some desire for creating/updating a tree widget from nested non-mapping iterables too (where we only care about hierarchy and values).

happy to leave that splitting for when someone needs that functionality

tlambert03:
thanks!

PR #161, opened 2023-04-21
[WIP] Simplify mouse drag simulation

andy-sweet:
Seems like the tests here can all be a bit flaky: https://github.com/pyapp-kit/superqt/actions/runs/4759738566/jobs/8459314640#step:8:510

It's not immediately obvious to me what the need for simulated fine-grained movements is in these tests, so I just made a single big move instead. Feel free to let me know if I'm missing an important detail here. Alternatively, we could use `waitSignal` within the for loop if each call is needed.

I'm still not entirely sure about the internals of qtbot and QTest (e.g. in particular where/what is the main event loop and what guarantees are made about processing events).

### DIFF
--- a/tests/test_sliders/test_slider.py
+++ b/tests/test_sliders/test_slider.py
@@ -56,10 +56,9 @@ def test_drag_handles(qtbot):
     assert rs._pressedIndex == 0
 
     # drag the left handle
-    with qtbot.waitSignals([rs.sliderMoved] * 13):  # couple less signals
-        for _ in range(15):
-            pos.setX(pos.x() + 2)
-            qtbot.mouseMove(rs, pos)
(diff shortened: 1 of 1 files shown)


andy-sweet:
Well that didn't work

tlambert03:
thank you for trying this here @andy-sweet  ... yeah, the flaky tests have definitely become annoying (they've only recently started failing, and it's definitely a big PITA at the moment).  I appreciate you taking a look.  I'm definitely ok with just skipping the offending tests for now (so we can get a release out without doing it manually).

most of the tests here were conceived to fill out test coverage (i.e. to make sure that any additional logic that I added on these sliders was actually getting tested somehow).  But, if doing so has now made testing on the whole broken, I'm happy to reconsider

PR #167, opened 2023-05-28
feat: Add `disable_throttling` fixture

Czaki:
This PR adds `disable_throttling` fixture to simplify test writing.

cc @kevinyamauchi

### DIFF
--- a/docs/utilities/throttling.md
+++ b/docs/utilities/throttling.md
@@ -44,3 +44,11 @@ For background on throttling and debouncing, see:
         docstring_style: numpy
         show_root_toc_entry: True
         show_root_heading: True
+
+
+## Throttling and tests.
+
+The throttlers are using internally timers to schedule proper calls. It is helpful
+in normal usage but may be problematic in tests. To avoid this problem, this package
+provides `disable_throttling` fixture that disables all throttlers and debouncers.
+The effect is that signals are emitted immediately and the timer is not used.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,6 +71,9 @@ font-fa6 = ["fonticon-fontawesome6"]
 font-mi6 = ["fonticon-materialdesignicons6"]
 font-mi7 = ["fonticon-materialdesignicons7"]
 
+[project.entry-points.pytest11]
+superqt = "superqt._pytest_plugin"
+
 [project.urls]
 Source = "https://github.com/pyapp-kit/superqt"
 Tracker = "https://github.com/pyapp-kit/superqt/issues"
--- a/src/superqt/_pytest_plugin.py
+++ b/src/superqt/_pytest_plugin.py
@@ -1,0 +1,17 @@
+import pytest
+
+from superqt.utils._throttler import GenericSignalThrottler
+
+
+def _throttle_mock(self):
+    self.triggered.emit()
+
+
+def _flush_mock(self):
+    """There are no waiting events."""
+
+
+@pytest.fixture()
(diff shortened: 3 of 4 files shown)


tlambert03 (commented):
Rather than a pytest plugin, how would you feel about a context manager?  (You can make it a pytest fixture internally at napari if you want to)

tlambert03:
I'd also be curious to see the context where this came up

Czaki:
It comes from here https://napari.zulipchat.com/#narrow/stream/212875-general/topic/flaky.20tests/near/361247624

A context manager is a little more complex as it needs to be applied to existing throttles when this PR apply patch before object creation.

tlambert03:
Couldn't the same be said for this fixture?  The fixture also gets invoked before object creation...

I ask only cause I don't love adding a new pytest entry point if it's avoidable (have run into rare cases where packages having pytest entry points has been an annoyance if I didn't care about them)

tlambert03:
The context manager would be just like the fixture monkeypatch, but importable, rather than discovered via the pytest plugin mechanism

Czaki:
I just worry that people will use such context managers in the wrong place (in code, or in test after object creation) 

But it could be done without pytest entry point. The user will need only to import it to the test file. Should I change it in this way?

tlambert03:
Hmm, I feel like that's also a little "surprising" (having to import a pytest fixture), so I'm torn.  

I don't worry too much about misusage (this is already a pretty advanced usage object, so it's not likely to see widespread usage, and if the test still fails, they're likely to immediately suspect they used it in the wrong place)

Czaki:
Changed to context manager

Czaki:
PySide6 test fails looks unrelated, but I'm not sure.

tlambert03:
> PySide6 test fails looks unrelated, but I'm not sure.

man, this is becoming a serious pain in the butt.  it seems like every release of pyside6 is breaking stuff.  i had it working last week, and indeed,  it looks like something (presumably the 6.5.1 release) broke it in the meantime.  😖

Czaki:
Yes. PySide6 is really annoying...

tlambert03:
thanks @Czaki ... i like this better for now.  One more question:  if the context manager accepted an object to actually disable (similar to `utils.signals_blocked()`) would that prevent you from doing what you need to?  It would be more "surgical" and make it far less likely that one sloppy user would mess up another user's throttles

Czaki:
Because of your worries, I prefer to create a fixture rather than a context manager. Providing a signal requires to use context manager after the object is created. That will be much more problematic in usage and maintenance.

tlambert03:
I'm receptive to going back to the fixture, but not sure I'm following you.  What's different here that makes it harder to do this on an object level instead of a class level?

tlambert03:
maybe you should just do this pattern in your own fixtures in napari?  Do you think this is such a common need that it should be offered as a pytest plugin?

Czaki:
Napari is really complex so it may happen there. But the problem with debouncer will happen in any code that uses it and is tested. 

But if a person does not use guards checking if all timers are ended during teardown he may and with random failures reported in unretalted tests. And for most users that ar not programmer in first place it will end with no testing code. Like Robert Haase in a few projects.

tlambert03:
As discussed on zulip, it was decided to implement this as a fixture in naparis tests.

Libraries needing this can use this pr to see how to disable throttling in their tests.  If multiple libraries find themselves needing it, we can reconsider a public api

PR #172, opened 2023-06-21
[proof of concept] make labelled range sliders expandable by modifying label range

alisterburt:
Hey Talley, I wanted a labelled QSlider where the label could be used to update the slider range like your labelled double slider. This PR is a minimal change that allows this to work for QLabeledSlider if the label range is manually updated

I've opened this PR so that we can discuss if there might be a better approach/worth adding some convenience for on the labelled slider class. If you think this is okay as-is then I'll add some tests before merge

```python
from qtpy.QtCore import Qt
from qtpy.QtWidgets import QApplication, QVBoxLayout, QWidget

from superqt import QLabeledSlider

app = QApplication([])

ORIENTATION = Qt.Orientation.Horizontal

w = QWidget()
w.setLayout(QVBoxLayout())

qls = QLabeledSlider(ORIENTATION)
qls.valueChanged.connect(lambda e: print("qls valueChanged", e))
qls.setRange(0, 100)
qls.setValue(30)

# update the label range - comment this line to revert to default behaviour
qls._label.setRange(0, 1000)

w.layout().addWidget(qls)

w.show()
app.exec_()
```

behaviour without this change (value can't be set outside of slider range (0, 100)):

https://github.com/pyapp-kit/superqt/assets/7307488/2bcc94f8-dc1f-4a89-9f31-9f901f04b8b3

behaviour with this change (range can be extended (value can be set to any value in label range (0, 1000), slider updates):

https://github.com/pyapp-kit/superqt/assets/7307488/e8cacde2-8aff-4ccf-b6d0-d4ba49c39caf

### DIFF
--- a/tests/test_sliders/test_labeled_slider.py
+++ b/tests/test_sliders/test_labeled_slider.py
@@ -85,3 +85,43 @@ def test_editing_float(qtbot):
     slider._label.setValue(0.5)
     slider._label.editingFinished.emit()
     assert slider.value() == 0.5
+
+
+@pytest.mark.parametrize("cls", [QLabeledSlider, QLabeledDoubleSlider])
+def test_extended_label_spinbox_range(cls, qtbot):
+    slider = cls()
+    mock = Mock()
+    qtbot.addWidget(slider)
+    slider.setRange(0, 10)
+    slider.rangeChanged.connect(mock)
+    assert slider._label.minimum() == 0
+    assert slider._label.maximum() == 10
+    assert slider.minimum() == 0
+    assert slider.maximum() == 10
+
+    # changing slider value without changing spinbox range should return a clipped value
+    # slider._label.setValue(20)
+    # slider._label.editingFinished.emit()  # simulate editing the label manually
+    # qtbot.wait(20)
+    # assert slider.value() == 10
+    # assert slider.maximum() == 10
+
+    # after changing label range, setting the label to a value outside the range should
+    # update the slider min/max as appropriate
+    slider.setLabelSpinBoxRange(-10, 20)
+
+    with qtbot.waitSignal(slider.rangeChanged):
+        slider._label.setValue(15)
+        assert slider._label.minimum() == -10
+        slider._label.editingFinished.emit()  # simulate editing the label manually
+        assert slider._label.minimum() == -10
+    assert slider.value() == 15
+    assert slider.minimum() == 0  # unchanged
+    assert slider.maximum() == 15  # changed
+
+    with qtbot.waitSignal(slider.rangeChanged):
+        slider._label.setValue(-5)
+        slider._label.editingFinished.emit()  # simulate editing the label manually
+    assert slider.value() == -5
+    assert slider.minimum() == -5  # changed
+    assert slider.maximum() == 15  # unchanged
--- a/src/superqt/sliders/_labeled.py
+++ b/src/superqt/sliders/_labeled.py
@@ -150,6 +150,9 @@ class QLabeledSlider(_SliderProxy, QAbstractSlider):
 
     def _setValue(self, value: float):
         """Convert the value from float to int before setting the slider value."""
+        value = int(value)
+        if self._slider.maximum() < value:
+            self._slider.setMaximum(value)
         self._slider.setValue(int(value))
 
     def _rename_signals(self):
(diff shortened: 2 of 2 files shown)


tlambert03:
Seems very reasonable...
@Czaki you expressed concern over in the other thread... what's your hesitation?

Czaki:
All Qt sliders, spinboxes, etc, do not allow for silent go out of range. This is really unexpected behavior. There is also no update of docs for that. 

So if anyone uses this widget as input, then they could write the following code that will assume that emitted value is in a given range. 

So at least, it should be mentioned in the documentation. 

Or a new class like `QLabeledSliderDynamic` should be added.

tlambert03:
You say silently, but couldn't we make sure that the proper signal is emitted?  

Most sliders don't have labels next to them in Qt, but with a label there, it's very similar to programmatically calling .setRange right?  (And that's fully supported)

Czaki:
Yes. But setRange is called from code. The user from GUI cannot change range.

tlambert03:
Perhaps we could add a flag to the init method to gate this feature?  Like extend_range_to_value

Czaki:
> Perhaps we could add a flag to the init method to gate this feature? Like extend_range_to_value

Yes. This is one of a possible solution.

tlambert03:
@alisterburt could you do that please?  Let's make it false by default so that this feature is opt in.  And then add a test to make sure that the rangeChanged (or whatever it is) signal is emitted when the feature is on and the label is changed, but not otherwise.

alisterburt:
Hi both, happy to gate in whatever way is best but it’s worth noting that with this implementation the behaviour is already gated:- unless the label (spinbox) range, not the slider range, is explicitly changed in code the user cannot specify a value outside the slider range in the label- if the label (spinbox) range is extended programmatically, the user can specify a value within that larger range and the slider will updateI can add an extra parameter to the init if we still feel it’s necessary but the current behaviour feels safe to me… what do we think?Sent from mobile - apologies for brevityOn 21 Jun 2023, at 12:13, Talley Lambert ***@***.***> wrote:﻿
@alisterburt could you do that please?  Let's make it false by default so that this feature is opt in.  And then add a test to make sure that the rangeChanged (or whatever it is) signal is emitted when the feature is on and the label is changed, but not otherwise.

tlambert03:
>unless the label (spinbox) range, not the slider range, is explicitly changed in code the user cannot specify a value outside the slider range in the label-

ohhhh, right!  Sorry, I missed the importance of that line in your example:

```py
# update the label range - comment this line to revert to default behaviour
qls._label.setRange(0, 1000)
```

> can add an extra parameter to the init if we still feel it’s necessary but the current behaviour feels safe to me… what do we think?

That said, I still think we need a somewhat public exposure of this.  I don't want napari, for example, accessing `._label.setRange` to "opt in" to this behavior... since `_label` may change in the future.

alisterburt:
cool! I'll cook something up now :)

tests/test_sliders/test_labeled_slider.py
```diff
@@ -85,3 +85,43 @@ def test_editing_float(qtbot):
...
+
+    # after changing label range, setting the label to a value outside the range should
+    # update the slider min/max as appropriate
+    slider.setLabelSpinBoxRange(-10, 20)
+
+    with qtbot.waitSignal(slider.rangeChanged):
+        slider._label.setValue(15)
+        assert slider._label.minimum() == -10
+        slider._label.editingFinished.emit()  # simulate editing the label manually
+        assert slider._label.minimum() == -10
```
alisterburt, line 117: something is going wrong here

`editingFinished` is the signal used to propagate label value changes down to the slider so I manually emit it for the test. This has the unintended side effect of changing `slider._label.minimum()`.

I tried to fix this by modifying 
https://github.com/pyapp-kit/superqt/blob/52d63ed68f6fcaaf0b4faab0254c41171b9b311c/src/superqt/sliders/_labeled.py#L593-L596

but the value for the label minimum has already been reset by the time we get here - I can't figure out where it has been reset...

alisterburt:
I had a go but got a little stuck, there is a side effect when emitting the `editingFinished` event which somehow resets `slider._label.minimum()` and `slider._label.maximum()` - I thought I had it pinned down (see comment in code above) but unfortunately that's not the culprit and I can't see where else this is being modified...

Would appreciate some help if possible - spent an hour or so bashing my head on this to no avail, happy to provide more context if needed

PR #177, opened 2023-08-05
refactor: make QEnumComboBox generic

tlambert03:


tlambert03:
eh, I forgot that pyside 6.5.0 and 6.5.1 were funny with Generic and that's why we did https://github.com/pyapp-kit/superqt/pull/164 ...
it looks like 6.5.2 works again.  but I'll stop short of pinning until we have more time to test it

PR #183, opened 2023-08-16
fix: fix parameter inspection on ensure_thread decorators

### LINKED ISSUE #182: bug: ensure_main_thread prevents proper signature evaluation
Just noticed that using the `@ensure_main_thread` decorator prevents Qt from knowing not to pass too many arguments.  (A nice thing about `signal.emit()` is that callbacks are supposed to be able to accept *up to* the number of arguments that the signal emits, but less should be ok).  with ensure_main_thread, the callback *must* accept all the arguments.

still looking into how to fix it

tlambert03:
fixes #182  ... and adds a test, but it sure isn't pretty at the moment :joy:

@Czaki, I'll keep playing with this (and digging into the qt source code) to see if I can come up with a more elegant way for Qt to be able to properly inspect the signature of a decorated wrapped function.  but if you have any other ideas i should try, feel free to propose!

**edit**
see also #185, which I think prefer now

### DIFF
--- a/src/superqt/utils/_ensure_thread.py
+++ b/src/superqt/utils/_ensure_thread.py
@@ -1,9 +1,11 @@
 # https://gist.github.com/FlorianRhiem/41a1ad9b694c14fb9ac3
 from __future__ import annotations
 
+import inspect
 from concurrent.futures import Future
 from functools import wraps
-from typing import TYPE_CHECKING, Callable, ClassVar, overload
+from textwrap import dedent
+from typing import TYPE_CHECKING, Any, Callable, ClassVar, overload
 
 from qtpy.QtCore import (
     QCoreApplication,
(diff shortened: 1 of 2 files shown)


tlambert03:
the crux of the pyside code is here: https://github.com/pyside/pyside-setup/blob/10941bc37450f2fe99cd3afa4e2331df0439d162/sources/pyside6/libpyside/pysidesignal.cpp#L1209C2-L1213C20

basically, it looks at `.__code__.co_argcount` and other `__code__` attributes. I tried patching those directly (without using compile and exec), but wasn't able to do so.  In any case, I'm not sure it would be any less black magic than the fix here.

I _do_ think this is important functionality to retain though (the ability to add more signals later without breaking APIs is a very nice thing in Qt/psygnal, and this is currently breaking it)

thoughts?

tlambert03:
linking https://github.com/pyapp-kit/psygnal/pull/228 here as well, since I suspect a similar phenomenon might be happening  in psygnal too.  basically, this will likely happen with *any* decorator that uses the pattern:

```
@wraps(func)
def inner(*args, **kwargs):
    func(*args, **kwargs)
```

since Qt doesn't know that `inner` has the same signature as `func`

PR #184, opened 2023-08-16
fix: fix callback of throttled/debounced decorated functions with mismatched args

tlambert03:
similar to #183, this fixes calling a qdebounced/qthrottled function with too many arguments

```python
class Emitter(QObject):
    sig = Signal(int, int, int)

@qdebounced
def func(a: int, b: int):
    ...

obj = Emitter()
obj.sig.connect(func)
obj.sig.emit(1, 2, 3)  # currently breaks, too many args
```

I had to do it manually here (meaning pass `func(*args[:max_args])`), because of a bug in pyside that prevents proper inspection of callable objects:
https://bugreports.qt.io/browse/PYSIDE-2423

This PR also changes the decorator pattern to return an instance of ThrottledCallable (whereas before it just returned an object that implemented the protocol).  I think it makes for a much clearer read


src/superqt/utils/_throttler.py
```diff
@@ -181,38 +181,55 @@ class QSignalDebouncer(GenericSignalThrottler):
...
 
-        def set_timeout(self, timeout: int) -> None:
-            ...
+        # even if we were to compile __call__ with a signature matching that of func,
+        # PySide wouldn't correctly inspect the signature of the ThrottledCallable
+        # instance: https://bugreports.qt.io/browse/PYSIDE-2423
+        # so we do it ourselfs and limit the number of positional arguments
+        # that we pass to func
+        try:
+            self._max_args: int | None = func.__code__.co_argcount
```
Czaki, line 215: Same remark as in #185.

tlambert03: updated to use get_max_args from #185

Czaki (approved):

src/superqt/utils/_throttler.py
```diff
@@ -221,28 +237,26 @@ def qthrottled(
...
     timer_type: Qt.TimerType = Qt.TimerType.PreciseTimer,
-) -> "ThrottledCallable[P, R]":
+) -> ThrottledCallable[P, R]:
     ...
 
 
 @overload
 def qthrottled(
-    func: Optional["Literal[None]"] = None,
+    func: Literal[None] | None = None,
```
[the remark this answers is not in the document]
tlambert03: looks like `func: None = ...` is also sufficient.  Will use that

PR #185, opened 2023-08-16
fix: fix parameter inspection on ensure_thread decorators (alternate)

### LINKED ISSUE #182: bug: ensure_main_thread prevents proper signature evaluation

tlambert03:
This is an alternative fix to #183, that doesn't use `compile` magic... and simply manually manages the `max_args` passed to a wrapped function.

cc @Czaki 

fixes #182


src/superqt/utils/_ensure_thread.py
```diff
@@ -166,17 +168,23 @@ def _run_in_thread(
...
-    *args,
-    **kwargs,
-):
+    args: tuple,
+    kwargs: dict,
+) -> Any:
+    try:
+        max_args: int | None = func.__code__.co_argcount
+    except AttributeError:
+        max_args = None
```
Czaki, line 177: this will not work properly:

```python
from functools import wraps

def func(a, b):
    return a + b

def wr(fun):
    @wraps(fun)
    def wr2(*args):
        return fun(*args) * 2
    return wr2

fn2 = wr(func)
```

then `fn2.__code__.co_argcount` will return 0.

This reveals two problems. At first this approach does not unwrap functions. 
At second the `*args` is not counted:

```python
>> def a(b, *c):
..    pass
>> a.__code__.co_argcount
1
```

Maybe use inspect (or reimplement most of logic, to handle partials etc)

Czaki (changes requested):
I think that there is a need to properly handle wrapped functions and `*args`

src/superqt/utils/_ensure_thread.py
tlambert03: thanks, yeah I was trying to avoid using inspect (so heavy...) but you're probably right.  lemme see if I can make a failing test then fix it

src/superqt/utils/_ensure_thread.py
```diff
@@ -166,17 +168,23 @@ def _run_in_thread(
...
+
     future = Future()  # type: ignore
     if thread is QThread.currentThread():
-        result = func(*args, **kwargs)
+        result = func(*args[:max_args], **kwargs)
```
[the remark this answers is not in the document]
tlambert03: :)

src/superqt/utils/_ensure_thread.py
tlambert03: yeah, I'm doing away with it, going back to inspect... but also moving where the inspection occurs higher up (to avoid doing it often)

tlambert03: actually. I think this one is correct by the way.  Signals include `self` in their connections... i think I tested that

tlambert03: ok updated... but all I really did was add an unwrap call:
```python
def _get_max_args(func: Callable) -> int | None:
     try:
         return inspect.unwrap(func).__code__.co_argcount  # type: ignore
     except Exception:
         return None
```

I added a number of additional tests, and I can't find a case where this simpler/faster approach fails.  so, if you still have issues with it, can you demonstrate a failure?

tlambert03: eh, never mind.  I changed it to:
```python
def _get_max_args(func: Callable) -> int | None:
    try:
        return len(inspect.signature(func).parameters)
    except Exception:
        return None
```

I do think the co_argcount is sufficient, and it's ~6µs instead of 50µs ... but it should only happen once (at decoration time), and i guess it's more likely to catch extreme cases.  so let's just play it safe

Czaki: I think that you should add these tests. It will be helpful if you try to speed up code.  (please notice that the first one is failing for current implementation):

```python
def test_get_max_args_simple():

    def fun1():
        pass

    assert _get_max_args(fun1) == 0

    def fun2(a):
        pass

    assert _get_max_args(fun2) == 1

    def fun3(a, b=1):
        pass

    assert _get_max_args(fun3) == 2

    def fun4(a, *, b=2):
        pass

    assert _get_max_args(fun4) == 1

    def fun5(a, *b):
        pass

    assert _get_max_args(fun5) is None



def test_get_max_args_wrapped():
    from functools import partial, wraps

    def fun1(a, b):
        pass

    assert _get_max_args(partial(fun1, 1)) == 1

    def dec(fun):
        @wraps(fun)
        def wrapper(*args, **kwargs):
            return fun(*args, **kwargs)
        return wrapper

    assert _get_max_args(dec(fun1)) == 2


def test_get_max_args_methods():

    class A:
        def fun1(self):
            pass

        def fun2(self, a):
            pass

        def __call__(self, a, b=1):
            pass


    assert _get_max_args(A().fun1) == 0
    assert _get_max_args(A().fun2) == 1
    assert _get_max_args(A()) == 2
```

tlambert03: ok, the following works for these cases:

```python
def get_max_args(func: Callable) -> int | None:
    """Return the maximum number of positional arguments that func can accept."""
    sig = signature(func)
    max_args = 0
    for param in sig.parameters.values():
        if param.kind == param.VAR_POSITIONAL:
            return None
        if param.kind in {param.POSITIONAL_ONLY, param.POSITIONAL_OR_KEYWORD}:
            max_args += 1
    return max_args
```

tlambert03: updated to use it here, along with your tests

src/superqt/utils/_util.py
```diff
@@ -0,0 +1,16 @@
+from __future__ import annotations
+
+from inspect import signature
+from typing import Callable
+
+
+def get_max_args(func: Callable) -> int | None:
+    """Return the maximum number of positional arguments that func can accept."""
+    sig = signature(func)
```
Czaki, line 9:
```suggestion
    try: 
	    sig = signature(func)
	except TypeError:
	    raise
	except:
	    return None
```

From documentation:

> Raises [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError) if no signature can be provided, and [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError) if that type of object is not supported. Also, if the annotations are stringized, and eval_str is not false, the eval() call(s) to un-stringize the annotations could potentially raise any kind of exception.

tlambert03: yeah, but i really don't think _this_ is the place where that error should be raised.  I think if we can't figure out how to limit the number of arguments, we just simply shouldn't try.  we shouldn't explicitly raise an exception

tlambert03: (i do agree it needed to be in a try block though!  I updated that )

tlambert03: actually... on second thought, I guess the current behavior of our decorators like `ensure_main_thread` would actually hide the proper error until emit time:

```python
# on main
In [1]: from qtpy.QtCore import QObject, Signal

In [2]: class Emitter(QObject):
   ...:     sig = Signal()
   ...:

In [3]: e = Emitter()

In [4]: from superqt.utils import ensure_main_thread

In [5]: e.sig.connect(ensure_main_thread(1))
Out[5]: <PyQt6.QtCore.QMetaObject.Connection object at 0x10f09ac00>

In [6]: e.sig.emit()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
File ~/dev/self/superqt/src/superqt/utils/_ensure_thread.py:93, in ensure_main_thread.<locals>._out_func.<locals>._func(*args, **kwargs)
     91 @wraps(func_)
     92 def _func(*args, **kwargs):
---> 93     return _run_in_thread(
     94         func_,
     95         QCoreApplication.instance().thread(),
     96         await_return,
     97         timeout,
     98         *args,
     99         **kwargs,
    100     )

File ~/dev/self/superqt/src/superqt/utils/_ensure_thread.py:174, in _run_in_thread(func, thread, await_return, timeout, *args, **kwargs)
    172 future = Future()  # type: ignore
    173 if thread is QThread.currentThread():
--> 174     result = func(*args, **kwargs)
    175     if not await_return:
    176         future.set_result(result)

TypeError: 'int' object is not callable
```

and this would catch that.  but I'm slightly wary of it because I've found cases before where `inspect.signature` returned a TypeError for actually callable objects (i.e. it could give us false positives)

tlambert03: yeah... for example:
```python
>>> inspect.signature(print)
  File "<stdin>", line 1, in <module>
  File "/Users/talley/micromamba/envs/py38/lib/python3.8/inspect.py", line 3105, in signature
    return Signature.from_callable(obj, follow_wrapped=follow_wrapped)
  File "/Users/talley/micromamba/envs/py38/lib/python3.8/inspect.py", line 2854, in from_callable
    return _signature_from_callable(obj, sigcls=cls,
  File "/Users/talley/micromamba/envs/py38/lib/python3.8/inspect.py", line 2308, in _signature_from_callable
    return _signature_from_builtin(sigcls, obj,
  File "/Users/talley/micromamba/envs/py38/lib/python3.8/inspect.py", line 2119, in _signature_from_builtin
    raise ValueError("no signature found for builtin {!r}".format(func))
ValueError: no signature found for builtin <built-in function print>
```

tests/test_utils.py
```diff
@@ -27,3 +28,64 @@ class Emitter(QObject):
...
+
+    def fun4(a, *, b=2):
+        pass
+
+    assert get_max_args(fun4) == 1
+
+    def fun5(a, *b):
+        pass
+
+    assert get_max_args(fun5) is None
```
Czaki, line 57:
```suggestion
    assert get_max_args(fun5) is None
    
    assert get_max_args(print) is None
```

as we know that print is valid test case.

Maybe also add test for raising on int?

tlambert03:
thanks again @Czaki ... are there still changes you are requesting?  (merging is blocked)

Czaki (approved):

PR #188, opened 2023-08-17
Implement throttling of methods

Czaki:
Allow to use `qdebounced` and `qthrottled` on methods with proper setting parent if use Qt objects.

### DIFF
--- a/src/superqt/utils/_throttler.py
+++ b/src/superqt/utils/_throttler.py
@@ -32,6 +32,7 @@ from concurrent.futures import Future
 from enum import IntFlag, auto
 from functools import wraps
 from typing import TYPE_CHECKING, Callable, Generic, TypeVar, overload
+from weakref import WeakKeyDictionary
 
 from qtpy.QtCore import QObject, Qt, QTimer, Signal
 
@@ -214,6 +215,8 @@ class ThrottledCallable(GenericSignalThrottler, Generic[P, R]):
         self.triggered.connect(self._set_future_result)
         self._name = None
 
+        self._obj_dkt = WeakKeyDictionary()
+
         # even if we were to compile __call__ with a signature matching that of func,
         # PySide wouldn't correctly inspect the signature of the ThrottledCallable
         # instance: https://bugreports.qt.io/browse/PYSIDE-2423
@@ -249,16 +252,22 @@ class ThrottledCallable(GenericSignalThrottler, Generic[P, R]):
         )
         throttler.setTimerType(self.timerType())
         throttler.setTimeout(self.timeout())
-        setattr(
-            obj,
-            self._name,
-            throttler,
-        )
+        try:
+            setattr(
+                obj,
+                self._name,
+                throttler,
+            )
+        except AttributeError:
+            self._obj_dkt[obj] = throttler
         return throttler
 
     def __get__(self, instance, owner):
         if instance is None or not self._name:
             return self
+
+        if instance in self._obj_dkt:
+            return self._obj_dkt[instance]
 
         parent = self.parent()
         if parent is None and isinstance(instance, QObject):
--- a/src/superqt/utils/_throttler.py
+++ b/src/superqt/utils/_throttler.py
@@ -214,7 +214,10 @@ class ThrottledCallable(GenericSignalThrottler, Generic[P, R]):
         # instance: https://bugreports.qt.io/browse/PYSIDE-2423
         # so we do it ourselfs and limit the number of positional arguments
         # that we pass to func
-        self._max_args: int | None = get_max_args(func)
+        if isinstance(func, staticmethod):
+            self._max_args = None
+        else:
+            self._max_args: int | None = get_max_args(func)
 
     def __call__(self, *args: P.args, **kwargs: P.kwargs) -> "Future[R]":  # noqa
         if not self._future.done():
@@ -235,6 +238,8 @@ class ThrottledCallable(GenericSignalThrottler, Generic[P, R]):
         self._name = name
 
     def __get__(self, instance, owner):
+        if isinstance(self.__wrapped__, staticmethod):
+            self._max_args = get_max_args(self.__wrapped__.__get__(instance, owner))
         if instance is None or not self._name:
             return self
         parent = self.parent()
--- a/src/superqt/utils/_throttler.py
+++ b/src/superqt/utils/_throttler.py
@@ -248,20 +248,26 @@ class ThrottledCallableDescriptor(ThrottledCallable):
     def __set_name__(self, owner, name):
         self._name = name
 
+    def _get_throttler(self, instance, owner, parent, obj):
+        throttler = ThrottledCallable(
+            self.__wrapped__.__get__(instance, owner),
+            self._kind,
+            self._emissionPolicy,
+            parent=parent,
+        )
+        throttler.setTimerType(self.timerType())
+        throttler.setTimeout(self.timeout())
+        setattr(
+            obj,
+            self._name,
+            throttler,
+        )
+        return throttler
+
     def __get__(self, instance, owner):
         parent = self.parent()
         if isinstance(self.__wrapped__, staticmethod):
-            setattr(
-                owner,
-                self._name,
-                ThrottledCallable(
-                    self.__wrapped__.__get__(instance, owner),
-                    self._kind,
-                    self._emissionPolicy,
-                    parent=parent,
-                ),
-            )
-            return getattr(owner, self._name)
+            return self._get_throttler(instance, owner, parent, owner)
 
         if instance is None or not self._name:
             return self
@@ -269,18 +275,7 @@ class ThrottledCallableDescriptor(ThrottledCallable):
         if parent is None and isinstance(instance, QObject):
             parent = instance
 
-        setattr(
-            instance,
-            self._name,
-            ThrottledCallable(
-                func=self.__wrapped__.__get__(instance, owner),
-                kind=self._kind,
-                emissionPolicy=self._emissionPolicy,
-                parent=parent,
-            ),
-        )
-
-        return getattr(instance, self._name)
+        return self._get_throttler(instance, owner, parent, instance)
 
 
 @overload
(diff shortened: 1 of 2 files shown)


src/superqt/utils/_throttler.py
```diff
@@ -227,16 +230,46 @@ def __call__(self, *args: P.args, **kwargs: P.kwargs) -> "Future[R]":  # noqa
...
+    def __get__(self, instance, owner):
+        if instance is None:
+            return self
+        parent = self.parent()
+        if parent is None and isinstance(instance, QObject):
+            parent = instance
+
+        setattr(
+            instance,
+            self._name,
```
tlambert03, line 253: type checker is pointing out that self._name could technically be `None`...  maybe raise an exception if we get there? (i've seen cases where `__set_name__` isn't called by various metaclasses... probably very rare, but might as well please the linter)

tlambert03 (approved):
very minor comment on an unlikely type error, then LGTM!

src/superqt/utils/_throttler.py
Czaki: I improve initial if to handle this.

src/superqt/utils/_throttler.py
```diff
@@ -227,16 +230,46 @@ def __call__(self, *args: P.args, **kwargs: P.kwargs) -> "Future[R]":  # noqa
         return self._future
 
     def _set_future_result(self):
-        result = self.__wrapped__(*self._args[: self._max_args], **self._kwargs)
+        if self._instance is not None:
+            result = self.__wrapped__(
+                self._instance, *self._args[: self._max_args], **self._kwargs
```
tlambert03, line 235: One more q:
Did you double check that max_args doesn't need to include instance here?  (I.e, do we not have an off by one issue?)

Czaki: Based on this test, the self should not be count: 
https://github.com/pyapp-kit/superqt/blob/504adf8bd0b8302b11f252222dddff099cc8eb4e/tests/test_utils.py#L80-L93

tlambert03: yeah, but max_args here is called on the unbound method (and that test is being called on a bound) method.  
It actually does fail:

```python
from qtpy.QtCore import QObject, Signal

from superqt.utils import qdebounced


class A(QObject):
    def __init__(self):
        super().__init__()
        self.count = 0

    @qdebounced(timeout=4)
    def callback(self, a):
        print(locals())


class E(QObject):
    signal = Signal(int, int)


app = QApplication([])
e = E()
a = A()
e.signal.connect(a.callback)
e.signal.emit(1, 2)
app.processEvents()

```

```
  File "/Users/talley/dev/self/superqt/src/superqt/utils/_throttler.py", line 234, in _set_future_result
    result = self.__wrapped__(
             ^^^^^^^^^^^^^^^^^
TypeError: A.callback() takes 2 positional arguments but 3 were given
zsh: abort      python x.py
```

I think you need to add 
```python
    def _set_future_result(self):
        if self._instance is not None:
            result = self.__wrapped__(
                *(self._instance, *self._args)[: self._max_args], **self._kwargs
            )
```

Czaki: Could you check my fix? It also handles staticmethod.

src/superqt/utils/_throttler.py
```diff
@@ -227,16 +230,50 @@ def __call__(self, *args: P.args, **kwargs: P.kwargs) -> "Future[R]":  # noqa
         return self._future
 
     def _set_future_result(self):
-        result = self.__wrapped__(*self._args[: self._max_args], **self._kwargs)
+        if self._instance is not None:
+            result = self.__wrapped__(
+                self._instance, *self._args[: self._max_args - 1], **self._kwargs
```
tlambert03, line 235: TypeError: max_args may be  `None`

src/superqt/utils/_throttler.py
tlambert03: also, looks like they changed staticmethod since py3.9... i think it might be more correct to swap the order of the decorators anyway?

src/superqt/utils/_throttler.py
```diff
@@ -256,11 +245,10 @@ def __get__(self, instance, owner):
             instance,
             self._name,
             ThrottledCallable(
-                func=self.__wrapped__,
+                func=self.__wrapped__.__get__(instance, owner),
```
tlambert03, line 248: love it

tests/test_throttler.py
```diff
@@ -26,6 +27,56 @@ def f2() -> str:
...
+        def __init__(self):
+            super().__init__()
+            self.count = 0
+
+        @qdebounced(timeout=4)
+        def callback(self):
+            self.count += 1
+
+        @qdebounced(timeout=4)
+        @staticmethod
```
tlambert03, line 64:
```suggestion
        @staticmethod
        @qdebounced(timeout=4)
```

i think this might fix the test error

tlambert03: (or at least, if it doesn't ... then it shows a good thing to test)

tlambert03: alternatively, since this doesn't *look* like it should have been an error on python3.8/3.9, you could try to figure out why staticmethods weren't `callable` back then, and special case it in get_max_args

tlambert03: yeah, I'm playing with it too... there's definitely something stange happening.  even if I pull out the actual function from the static method at `func.__func__`, its *still* getting reassigned back to `__wrapped__` at some later point, and causing an error (on py38)

Czaki: Could you paste the code snippet that reproduces the problem?

tlambert03: i'm just running your `test_debouncer_method_definition` test.  i think i have a fix... just a min

tlambert03: oh I see you posted a fix as well... hang on

Czaki: But how it pass on CI and fail on your machine?

tlambert03: the issue, by the way, was that `__set_name__` appears to be *reassigning* `self.__wrapped__`.

>But how it pass on CI and fail on your machine?

because I hadn't pulled your new variant with a new class

tlambert03:
this is fine, since all the tests pass, but it's pretty verbose, and I think it can be done without special casing staticmethod everywhere, and without the new descriptor.  feel free to be done with it though.  if I can't come up with a cleaner approach, i'll merge it soon

src/superqt/utils/_throttler.py
```diff
@@ -231,12 +234,62 @@ def _set_future_result(self):
...
+    def __set_name__(self, owner, name):
+        self._name = name
+
+    def __get__(self, instance, owner):
+        parent = self.parent()
+        if isinstance(self.__wrapped__, staticmethod):
+            setattr(
+                owner,
+                self._name,
+                ThrottledCallable(
```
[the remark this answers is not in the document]
Czaki: Yes. I restore this information and simplified a little logic.

tlambert03:
my  suggestions are here: https://github.com/Czaki/superqt/pull/3 
but i like what you did with it in the menatime

src/superqt/utils/_throttler.py
```diff
@@ -354,10 +405,16 @@ def _make_decorator(
...
+
+        instance: object | None = getattr(func, "__self__", None)
+        if isinstance(instance, QObject) and parent is None:
+            parent = instance
         policy = EmissionPolicy.Leading if leading else EmissionPolicy.Trailing
-        obj = ThrottledCallable(func, kind, policy)
+        obj = ThrottledCallable(func, kind, policy, parent=parent)
         obj.setTimerType(timer_type)
         obj.setTimeout(timeout)
         return wraps(func)(obj)
```
Czaki, line 420: this line overwrites `__wrapped__`

Czaki:
I inspire your proposition but also found the reason of overwritten of `__wprapped__`. What You think about current version?

tlambert03:
works for me!  I think the only remaining thing is the [setattr](https://github.com/pyapp-kit/superqt/pull/188/files#diff-2380b7357c937931633e333f7791baabf19de112168556af51e00f82234a5d4cR252-R256) ... which could potentially cause AttributeErrors in rare cases (and it's ok if it fails)

Czaki:
> works for me! I think the only remaining thing is the [setattr](https://github.com/pyapp-kit/superqt/pull/188/files#diff-2380b7357c937931633e333f7791baabf19de112168556af51e00f82234a5d4cR252-R256) ... which could potentially cause AttributeErrors in rare cases (and it's ok if it fails)

It is not ok, as it means separate throttle per call. Even if it not a problem when connecting to a signal, this is a problem when someone calls such a function directly.

Czaki:
Did you know the exact scenario when assignment may fail?

tlambert03:
I'm not saying not to use it, I'm saying to put it in a try catch.  Since if you can't set the attribute you're screwed anyway (so you might as well fall back to a new throttle on each access). 

This will fail in all cases where the object doesn't have `__dict__`, such as any c extension (mypyc or other), or cases with slots.  I did exactly what you're doing here in other libraries and ended up having to put in the try except later, so please add it.  It won't hurt

Czaki:
added WeakKeyDictionary as fallback

tlambert03:
> added WeakKeyDictionary as fallback

that'll work!  thanks :)

tlambert03 (approved):
good stuff!

Czaki:
```python
def test_class_with_slots(qtbot):
    class A:
        __slots__ = ("count",)
        def __init__(self):
            self.count = 0

        @qdebounced(timeout=4)
        def callback(self):
            self.count += 1

    a = A()
    for _ in range(10):
        a.callback()

    qtbot.wait(5)
    assert a.count == 1
```
ends with 

```
E       TypeError: cannot create weak reference to 'A' object
```

tlambert03:
lol :)

Czaki:
This fix it:

```python
def test_class_with_slots(qtbot):
    class A:
        __slots__ = ("count", "__weakref__")
        def __init__(self):
            self.count = 0

        @qdebounced(timeout=4)
        def callback(self):
            self.count += 1

    a = A()
    for _ in range(10):
        a.callback()

    qtbot.wait(5)
    assert a.count == 1
```
But it may need to add proper docstring/error message.

tlambert03:
yeah, exactly, someone would need to opt in to make their slotted object weakreferencable (or use the `id()` trick)

Czaki:
And if do not want, then still could use `connect(throttled(self.method))`

tlambert03:
yep exactly.  So, leave it?

PR #189, opened 2023-08-18
fix: Add descriptive exception when fail to add instance to weakref dictionary 

Czaki:


src/superqt/utils/_throttler.py
```diff
@@ -259,7 +259,16 @@ def _get_throttler(self, instance, owner, parent, obj):
...
-            self._obj_dkt[obj] = throttler
+            try:
+                self._obj_dkt[obj] = throttler
+            except TypeError as e:
+                raise TypeError(
+                    "To use qthrottled or qdebounced as a class decorator, "
+                    "there is a need to have __dict__ or allow weak references"
+                    " to the class instance, that requires __weakref__ to "
+                    "be added to __slots__. If you don't want to add then "
+                    "qthrottled or qdebounced can be used as a function."
```
tlambert03, line 270:
```suggestion
                    "To use qthrottled or qdebounced as a method decorator, "
                    "objects must have  `__dict__` or be weak referenceable. "
                    "Please either add `__weakref__` to `__slots__` or use"
                    "qthrottled/qdebounced as a function (not a decorator)."
```

tlambert03:
now I think we've got it covered for sure!  thanks! :)

PR #191, opened 2023-08-21
feat: add error `exceptions_as_dialog` context manager to catch and show Exceptions

tlambert03:
```python

from superqt.utils import exceptions_as_dialog

app = QApplication([])

with exceptions_as_dialog():
    raise Exception("This will be caught and shown in a QMessageBox")

# with exceptions_as_dialog(ValueError):
#     1 / 0  # ZeroDivisionError is not caught, so this will raise

with exceptions_as_dialog(msg_template="Error: {exc_value}"):
    raise Exception("This message will be used as 'exc_value'")

for _i in range(3):
    with exceptions_as_dialog(AssertionError, use_global_err_message=True):
        raise AssertionError("Uncheck the checkbox to ignore this in the future")

app.exec()
```

### DIFF
--- a/src/superqt/utils/_errormsg_context.py
+++ b/src/superqt/utils/_errormsg_context.py
@@ -29,9 +29,11 @@ class exceptions_as_dialog(AbstractContextManager):
     msg_template : str, optional
         The message to show in the `QMessageBox`. The message will be formatted
         using three variables:
+
         - `exc_value`: the exception instance
         - `exc_type`: the exception type
         - `tb`: the traceback as a string
+
         The default template is the content of the exception: `"{exc_value}"`
     buttons : QMessageBox.StandardButton, optional
         The buttons to show in the `QMessageBox`, by default
--- a/src/superqt/utils/_errormsg_context.py
+++ b/src/superqt/utils/_errormsg_context.py
@@ -1,0 +1,85 @@
+import traceback
+from contextlib import AbstractContextManager
+from types import TracebackType
+
+from qtpy.QtWidgets import QErrorMessage, QMessageBox, QWidget
+
+_GLOBAL_ERR = None
+
+
+def global_err_message() -> QErrorMessage:
+    global _GLOBAL_ERR
+    if _GLOBAL_ERR is None:
+        _GLOBAL_ERR = QErrorMessage()
+    return _GLOBAL_ERR
+
+
+class exceptions_as_dialog(AbstractContextManager):
+    """Context manager that shows a QMessageBox when an exception is raised.
+
+    Example
+    -------
+    ```python
+    with exceptions_as_dialog():
+        raise Exception("This will be caught and shown in a QMessageBox")
+
+    with exceptions_as_dialog(ValueError):
+        1 / 0  # ZeroDivisionError is not caught, so this will raise
+
+    with exceptions_as_dialog(msg_template="Error: {exc_value}"):
+        raise Exception("This message will be used as 'exc_value'")
+
+    for _i in range(3):
+        with exceptions_as_dialog(AssertionError, global_err_message=True):
+            assert False, "Uncheck the checkbox to ignore this in the future"
+    ```
+    """
+
+    widget: QMessageBox | None
+    raised: bool
+
+    def __init__(
+        self,
+        exceptions: type[BaseException] | tuple[type[BaseException], ...] = (Exception),
+        icon: QMessageBox.Icon = QMessageBox.Icon.Critical,
+        title: str = "An error occurred",
+        msg_template: str = "{exc_value}",
+        buttons: QMessageBox.StandardButton = QMessageBox.StandardButton.Ok,
+        parent: QWidget | None = None,
+        use_global_err_message: bool = False,
+    ):
+        self.exceptions = exceptions
+        self.msg_template = msg_template
+        self.use_global_err_message = use_global_err_message
+        self.raised = False
+        if not use_global_err_message:
+            self.widget = QMessageBox(icon, title, "", buttons, parent)
+        else:
+            self.widget = None
+
+    def __enter__(self) -> "exceptions_as_dialog":
+        return self
+
+    def __exit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc_value: BaseException | None,
+        tb: TracebackType | None,
+    ) -> bool:
+        if not (exc_value is not None and isinstance(exc_value, self.exceptions)):
+            return False  # let it propagate
+
+        self.raised = True
+        if "tb" in self.msg_template:
+            _tb = "\n".join(traceback.format_exception(exc_type, exc_value, tb))
+        else:
+            _tb = ""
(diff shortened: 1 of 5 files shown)


src/superqt/utils/_errormsg_context.py
```diff
@@ -0,0 +1,87 @@
...
+    ) -> bool:
+        if not (exc_value is not None and isinstance(exc_value, self.exceptions)):
+            return False  # let it propagate
+
+        self.raised = True
+        if "tb" in self.msg_template:
+            _tb = "\n".join(traceback.format_exception(exc_type, exc_value, tb))
+        else:
+            _tb = ""
+        text = self.msg_template.format(exc_value=exc_value, exc_type=exc_type, tb=_tb)
```
Czaki, line 79:
```suggestion
        if "{tb}" in self.msg_template:
            _tb = "\n".join(traceback.format_exception(exc_type, exc_value, tb))
        else:
            _tb = ""
        text = self.msg_template.format(exc_value=exc_value, exc_type=exc_type, tb=_tb)
```

Czaki (commented):
looks ok 
It may be nice to create a local error message (share in a given context, not globally).

src/superqt/utils/_errormsg_context.py
Czaki: I mean that one may provide instance of `QErrorMessage` as an argument to decide about scope.

tlambert03: ah yep, I'll do that.  I also learned that `QErrorMessage.qtHandler()` is a static method to get a global instance... so I don't have to create one

tlambert03:
ok, ready for final review

src/superqt/utils/_errormsg_context.py
```diff
@@ -0,0 +1,145 @@
...
+    parent : QWidget | None, optional
+        The parent widget of the `QMessageBox`, by default `None`
+    use_error_message : bool | QErrorMessage, optional
+        Whether to use a `QErrorMessage` instead of a `QMessageBox`. By default
+        `False`. `QErrorMessage` shows a checkbox that the user can check to
+        prevent seeing the message again (based on the text of the formatted
+        `msg_template`.) If `True`, the global `QMessageError.qtHandler()`
+        instance is used to maintain a history of dismissed messages. You may also pass
+        a `QErrorMessage` instance to use a specific instance. If `use_error_message` is
+        True, `parent` is ignored.
```
Czaki, line 48: "Is not false?" As parent, it is also ignored when provided QErrorMessage? 

How it integrates with the change in global style (like in napari)?

src/superqt/utils/_errormsg_context.py
```diff
@@ -0,0 +1,145 @@
...
+    widget: QMessageBox | None
+    exception: BaseException | None
+
+    def __init__(
+        self,
+        exceptions: type[BaseException] | tuple[type[BaseException], ...] = Exception,
+        icon: QMessageBox.Icon = QMessageBox.Icon.Critical,
+        title: str = "An error occurred",
+        msg_template: str = "{exc_value}",
+        buttons: QMessageBox.StandardButton = QMessageBox.StandardButton.Ok,
```
Czaki, line 96: If we would like to allow select buttons, there should be an option to check what users select. So save the `self.widget.exec()` result.

tests/test_utils.py
```diff
@@ -91,3 +93,35 @@ def __call__(self, a, b=1):
...
+    with exceptions_as_dialog(use_error_message=err):
+        QTimer.singleShot(0, err.accept)
+        raise AssertionError("Uncheck the checkbox to ignore this in the future")
+
+    # tb formatting smoke test, and return value checking
+    exc = ValueError("Bad Val")
+    with exceptions_as_dialog(msg_template="{tb}") as ctx:
+        qtbot.addWidget(ctx.widget)
+        QTimer.singleShot(100, accept)
+        raise exc
```
Czaki, line 125: It will be nice to test messages somehow. But not important.

src/superqt/utils/_errormsg_context.py
tlambert03: yeah, i had made a little note in the `__exit__` method to that effect.   Do you remember whether the  Dialog itself stores the last result of exec?  Or not in the case of multiple buttons?  I can just add another attribute

src/superqt/utils/_errormsg_context.py
tlambert03: > As parent, it is also ignored when provided QErrorMessage?

Yes, ignored there as well.  If someone is handling the construction of the QErrorMessage instance, I think they should also handle passing the parent to the constructor (and we shouldn't override it).

> How it integrates with the change in global style (like in napari)?

 at least on macos, QErrorMessage actually uses a native dialog, anyway, so it can't be styled:

```
from qtpy.QtWidgets import QApplication, QErrorMessage

app = QApplication([])
err = QErrorMessage()
err.setStyleSheet("QErrorMessage { background-color: red }")
err.showMessage("This is an error message")
app.exec_()

```

Screenshot 2023-08-21 at 11 28 25 AM

src/superqt/utils/_errormsg_context.py
Czaki: `dialog.result()`

tlambert03: ah, ok... so, what do you think?

```python
with exceptions_as_dialog() as ctx:
    ...
ctx.dialog.result()
```

or 
```python
with exceptions_as_dialog() as ctx:
    ...
ctx.exec_result()
```

tlambert03: i think i prefer the former... it would also allow `ctx.dialog.clickedButton()` for example

src/superqt/utils/_errormsg_context.py
Czaki: Ok. Stil docstring need update (from my point of understanding)

src/superqt/utils/_errormsg_context.py
```diff
@@ -0,0 +1,148 @@
...
+        `msg_template`.) If `True`, the global `QMessageError.qtHandler()`
+        instance is used to maintain a history of dismissed messages. You may also pass
+        a `QErrorMessage` instance to use a specific instance. If `use_error_message` is
+        True, `parent` is ignored.
+
+    Attributes
+    ----------
+    dialog : QMessageBox | None
+        The `QMessageBox` instance that was created. If `use_error_message` is
+        True, this will be `None`.
```
Czaki, line 56: Here also condition requires update.

Maybe mention how to get the result?

Czaki (approved):
Looks good. I think that docstrings should be updated, but you are a native speaker.

tlambert03:
> I think that docstrings should be updated, but you are a native speaker.

can you elaborate?  Happy to change anything you think was unclear... was it just the one comment you left?  Or something you didn't leave?

src/superqt/utils/_errormsg_context.py
tlambert03: changed to Is not False. 👍

PR #205, opened 2023-09-23
Fix IntEnum for python 3.11

### LINKED ISSUE #202: QEnumComboBox _get_name fails on IntEnum
@Czaki, [this issue](https://github.com/pymmcore-plus/napari-micromanager/issues/286) was posted over in napari-micromanager, but it's actually a bug in the QEnumComboBox as used by https://github.com/jacopoabramo/napari-live-recording 

It appears that the `_get_name` strategy doesn't work for things like IntEnums (and probably other enum types):

```python
from enum import IntEnum

from superqt import QEnumComboBox


class MyEnum(IntEnum):
    ONE = 1

app = QApplication([])
cb = QEnumComboBox(enum_class=MyEnum)
```

causes 

```pytb
  File "/Users/talley/dev/self/pymmcore-widgets/x.py", line 11, in <module>
    cb = QEnumComboBox(enum_class=MyEnum)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/talley/dev/self/superqt/src/superqt/combobox/_enum_combobox.py", line 42, in __init__
    self.setEnumClass(enum_class, allow_none)
  File "/Users/talley/dev/self/superqt/src/superqt/combobox/_enum_combobox.py", line 53, in setEnumClass
    _names = dict.fromkeys(names)  # remove duplicates/aliases, keep order
             ^^^^^^^^^^^^^^^^^^^^
  File "/Users/talley/dev/self/superqt/src/superqt/combobox/_enum_combobox.py", line 16, in _get_name
    enum_value.__str__.__module__ != "enum"
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'method-wrapper' object has no attribute '__module__'. Did you mean: '__reduce__'?
```

Czaki:
Closes #202 

The error described in #202 is Python 3.11 specific. Caused by this change:

> Changed in version 3.11: [__str__()](https://docs.python.org/3/reference/datamodel.html#object.__str__) is now int.__str__() to better support the replacement of existing constants use-case. [__format__()](https://docs.python.org/3/reference/datamodel.html#object.__format__) was already int.__format__() for that same reason.

Described in [documentation](https://docs.python.org/3/library/enum.html#enum.IntEnum). 

I have changed the `_get_name` strategy to assume that if we cannot determine the `__str__` module, then we could assume that it is not overwritten by the user and use the default strategy. 

### Summary by CodeRabbit

- Bug Fix: Enhanced the `_get_name` function in `src/superqt/combobox/_enum_combobox.py` to handle cases where the `__str__` method of the `enum_value` object is overridden or belongs to a different module. This ensures more robust and reliable functionality.
- Test: Added a new test case `test_simple_create_int_enum` in `tests/test_enum_comb_box.py` to verify the correct creation of a `QEnumComboBox` with an `IntEnum1` class, improving the overall test coverage.

### DIFF
--- a/src/superqt/combobox/_enum_combobox.py
+++ b/src/superqt/combobox/_enum_combobox.py
@@ -12,10 +12,8 @@ NONE_STRING = "----"
 
 def _get_name(enum_value: Enum):
     """Create human readable name if user does not implement `__str__`."""
-    if (
-        enum_value.__str__.__module__ != "enum"
-        and not enum_value.__str__.__module__.startswith("shibokensupport")
-    ):
+    str_module = getattr(enum_value.__str__, "__module__", "enum")
+    if str_module != "enum" and not str_module.startswith("shibokensupport"):
         # check if function was overloaded
         name = str(enum_value)
     else:
--- a/tests/test_enum_comb_box.py
+++ b/tests/test_enum_comb_box.py
@@ -1,4 +1,4 @@
-from enum import Enum
+from enum import Enum, IntEnum
 
 import pytest
 
@@ -34,6 +34,12 @@ class Enum4(Enum):
     a_1 = 1
     b_2 = 2
     c_3 = 3
+
+
+class IntEnum1(IntEnum):
+    a = 1
+    b = 2
+    c = 5
 
 
 def test_simple_create(qtbot):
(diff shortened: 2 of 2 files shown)


tlambert03:
Thank you!  And thanks also for the explanation

PR #207, opened 2023-09-25
Add support for flag enum

Czaki:
As followup #205 I have checked that we do not have a test for another enum base class. So I prevent future failures, I added tests and spotted that flag support was not fully functional as it does not support element combinations. 

Also add `StrEnum` test. 

### Summary by CodeRabbit

- New Feature: Enhanced the `QEnumComboBox` class to support `Flag` enums in addition to regular enums. This allows users to select multiple options from a combo box.
- Refactor: Improved the internal handling of enum classes in `QEnumComboBox` by introducing helper functions for retrieving enum member names and values.
- Test: Added comprehensive test cases for the new `Flag` enum support in `QEnumComboBox`, ensuring robust functionality and error handling.

### DIFF
--- a/src/superqt/combobox/_enum_combobox.py
+++ b/src/superqt/combobox/_enum_combobox.py
@@ -1,3 +1,4 @@
+import sys
 from enum import Enum, EnumMeta, Flag
 from functools import reduce
 from itertools import combinations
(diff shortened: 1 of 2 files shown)


tlambert03:
Thanks @Czaki 

Looks like it needs updating for everything below 3.11

Czaki:
> Looks like it needs updating for everything below 3.11

Yes. I also hit problems as there was a bug in the new name mechanism in Python 3.11(.3). 

As it is fixed in 3.11.5 then I left only the workaround code with a docstring explanation.

tlambert03 (approved):
great digging.  thanks for the updates and all the great tests!  LGTM

PR #209, opened 2023-09-30
feat: add QIcon backed by iconify

tlambert03:
[Iconify](https://icon-sets.iconify.design) is a pretty great API unifying pretty much all of the major font-icon libraries into a single API from which you can request standardized SVGs.  I wrapped the API in library [`pyconify`](https://github.com/pyapp-kit/pyconify) that makes it easier to work with in python, and caches all svgs locally for faster/offline usage (after first using online once).  This PR makes a simple QIcon wrapper:

```python
from qtpy.QtCore import QSize
from qtpy.QtWidgets import QApplication, QPushButton

from superqt import QIconifyIcon

app = QApplication([])

btn = QPushButton()
# search https://icon-sets.iconify.design for available icon keys
btn.setIcon(QIconifyIcon("fa:smile-o", color="blue"))
btn.setIconSize(QSize(60, 60))
btn.show()

app.exec()
```

A downside of using this is that it's SVG based, which makes it a bit harder than font libraries to change colors simply using style sheets (like napari does for its themes... though even that may be possible).  A big upside is that you no longer need to include the whole font-family file just to grab a couple icons, and you can pick and choose from all of the libraries without having to bring in a new plugin for each one.  Also, for reasons beyond just color changing, SVGs are preferable in many cases to fonts (animations are possible, for example)

### DIFF
--- a/src/superqt/iconify/__init__.py
+++ b/src/superqt/iconify/__init__.py
@@ -34,15 +34,13 @@ class QIconifyIcon(QIcon):
         Icon set prefix and name. May be passed as a single string in the format
         `"prefix:name"` or as two separate strings: `'prefix', 'name'`.
     color : str, optional
-        Icon color. Replaces currentColor with specific color, resulting in icon with
-        hardcoded palette.
+        Icon color. If not provided, the icon will appear black (the icon fill color
+        will be set to the string "currentColor").
     flip : str, optional
-        Flip icon.
+        Flip icon.  Must be one of "horizontal", "vertical", "horizontal,vertical"
     rotate : str | int, optional
-        Rotate icon. If an integer is provided, it is assumed to be in degrees.
-    prefix : str, optional
-        If not None, the temp file name will begin with that prefix, otherwise a default
-        prefix is used.
+        Rotate icon. Must be one of 90, 180, 270,
+        or 1, 2, 3 (equivalent to 90, 180, 270, respectively)
     dir : str, optional
         If 'dir' is not None, the file will be created in that directory, otherwise a
         default directory is used.
@@ -62,7 +60,6 @@ class QIconifyIcon(QIcon):
         color: str | None = None,
         flip: Flip | None = None,
         rotate: Rotation | None = None,
-        prefix: str | None = None,
         dir: str | None = None,
     ):
         try:
@@ -72,7 +69,5 @@ class QIconifyIcon(QIcon):
                 "pyconify is required to use QIconifyIcon. "
                 "Please install it with `pip install pyconify`"
             ) from None
-        self.path = temp_svg(
-            *key, color=color, flip=flip, rotate=rotate, prefix=prefix, dir=dir
-        )
+        self.path = temp_svg(*key, color=color, flip=flip, rotate=rotate, dir=dir)
         super().__init__(self.path)
--- a/src/superqt/iconify/__init__.py
+++ b/src/superqt/iconify/__init__.py
@@ -65,7 +65,7 @@ class QIconifyIcon(QIcon):
     ):
         try:
             from pyconify import svg_path
-        except ModuleNotFoundError as e:
+        except ModuleNotFoundError as e:  # pragma: no cover
             raise ImportError(
                 "pyconify is required to use QIconifyIcon. "
                 "Please install it with `pip install pyconify` or use the "
--- a/src/superqt/iconify/__init__.py
+++ b/src/superqt/iconify/__init__.py
@@ -22,7 +22,7 @@ class QIconifyIcon(QIcon):
     icon:  `QIconifyIcon("bi:bell")`.
 
     This class is a thin wrapper around the
-    [pyconify](https://github.com/pyapp-kit/pyconify) `temp_svg` function. It pulls SVGs
+    [pyconify](https://github.com/pyapp-kit/pyconify) `svg_path` function. It pulls SVGs
     from iconify, creates a temporary SVG file and uses it as the source for a QIcon.
     SVGs are cached to disk, and persist across sessions (until `pyconify.clear_cache()`
     is called).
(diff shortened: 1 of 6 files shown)


tests/test_iconify.py
```diff
@@ -0,0 +1,14 @@
+import pytest
+from qtpy.QtWidgets import QPushButton
+
+from superqt import QIconifyIcon
+
+
+def test_qiconify(qtbot):
+    pytest.importorskip("pyconify")
+    icon = QIconifyIcon("bi:alarm-fill", color="red", rotate=90)
```
Czaki, line 9:
```suggestion
def test_qiconify(qtbot, tmp_path):
    pytest.importorskip("pyconify")
    icon = QIconifyIcon("bi:alarm-fill", color="red", rotate=90, dir=str(tmp-path))
```

keep test artifacts in test dirs

src/superqt/iconify/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+        Icon color. If not provided, the icon will appear black (the icon fill color
+        will be set to the string "currentColor").
+    flip : str, optional
+        Flip icon.  Must be one of "horizontal", "vertical", "horizontal,vertical"
+    rotate : str | int, optional
+        Rotate icon. Must be one of 90, 180, 270,
+        or 1, 2, 3 (equivalent to 90, 180, 270, respectively)
+    dir : str, optional
+        If 'dir' is not None, the file will be created in that directory, otherwise a
+        default directory is used.
```
Czaki, line 45:
```suggestion
    dir : str, optional
        If 'dir' is not None, the file will be created in that directory, otherwise a
        default [directory](https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp) is used.
```

Czaki (commented):
It looks ok. Few minor remarks.

Did you plan to ship pyconify to conda?|

Why not allow to cache icons svg between sessions?

tlambert03:
> Why not allow to cache icons svg between sessions?

They are cached between sessions:
https://github.com/pyapp-kit/pyconify#cache

The main difference between temp_svg and svg is that temp_svg returns a file path and svg returns bytes.  But they both use the same persistent cache

yes I’ll put it on conda

tests/test_iconify.py
tlambert03: Actually we’ll need to mock the pyconify cache here too

Czaki:
Hm. Did you try to play with `QIconEngine`? Qt documentation suggest that it could be used to render svg without need to load it from drive.

tlambert03:
> Did you try to play with QIconEngine? Qt documentation suggest that it could be used to render svg without need to load it from drive.

Yep, I did quite a bit.  I originally started with a similar strategy to what i did in napari: https://github.com/napari/napari/blob/main/napari/_qt/qt_resources/_svg.py

However, after looking into more in this round, I'm less of a fan of that approach.  If you look into the c++ code of the builtin QIconEngine, you'll find that it actually does quite a bit more than a naive approach might, it handles caching of pixmaps for various sizes, and better handles rendering svgs for different states (enabled/disabled, active/inactive).  So, after a fair amount of experimentation, I decided that simply writing a file to disk and letting the builtin QIconEngine handle it is well worth the cost.  It still all loads in about 10µs (after the original caching from the API), so it's hardly a performance limitation.  And there's a large amount of code that no longer needs to be maintained.  So that's why I went with this approach.

I will, however, update pyconify to be able to give you the path to the actual cached file in the pyconify cache (rather than this current approach, which costs one additional local file copy from the pyconify cache to the tempdir)

tlambert03:
@Czaki, will likely merge this soon.  Would welcome your thoughts on `QIconEngine`, specifically, I'm curious what specifically you expect/hope would be gained by using it.  with the current implementation, there is only one disk read (from the pyconify) cache.  So I expect it will be both performant *and* take advantage of all of the extra goodies provided by the [default QIconEngine](https://github.com/qt/qtbase/blob/dev/src/gui/image/qicon.cpp#L130) (not to mention it's less code to maintain)

tests/test_iconify.py
```diff
@@ -0,0 +1,22 @@
...
+from qtpy.QtWidgets import QPushButton
+
+from superqt import QIconifyIcon
+
+if TYPE_CHECKING:
+    from pytestqt.qtbot import QtBot
+
+
+def test_qiconify(qtbot: "QtBot", monkeypatch: "pytest.MonkeyPatch") -> None:
+    monkeypatch.setenv("PYCONIFY_CACHE", "0")
```
Czaki, line 13:
```suggestion
def test_qiconify(qtbot: "QtBot", monkeypatch: "pytest.MonkeyPatch", tmp_path: "pathlib.Path") -> None:
    monkeypatch.setenv("PYCONIFY_CACHE", str(tmp_path))
```

Instead of depending on pyconify clean mechanism just store cache in test managed directory?

Czaki (approved):
As far as I check the `svg_path` is improperly uses cache. I have created PR to fix this: https://github.com/pyapp-kit/pyconify/pull/5

I have two small remarks (one typo, and one for test). But general image is good.

src/superqt/iconify/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+        flip: Flip | None = None,
+        rotate: Rotation | None = None,
+        dir: str | None = None,
+    ):
+        try:
+            from pyconify import svg_path
+        except ModuleNotFoundError as e:
+            raise ImportError(
+                "pyconify is required to use QIconifyIcon. "
+                "Please install it with `pip install iconify` or use the "
```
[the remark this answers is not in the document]
tlambert03: yeah, i think it's verbose.  if someone wants conda i think they generally know how to check, and it's also in the readme of pyconify

tlambert03:
thanks @Czaki!

merged your PR, and bumped the dependency here.

PR #211, opened 2023-10-08
fix: don't enqueue Worker for deletion in `new_worker_qthread`

tlambert03:
on a tip from @andy-sweet ... to better match the pattern shown in https://doc.qt.io/qt-6/qthread.html#details

@andy-sweet, let me know if you agree this is all that is needed

### DIFF
--- a/src/superqt/utils/_qthreading.py
+++ b/src/superqt/utils/_qthreading.py
@@ -881,6 +881,7 @@ def new_worker_qthread(
     worker.moveToThread(thread)
     thread.started.connect(worker.work)
     worker.finished.connect(thread.quit)
+    thread.finished.connect(worker.deleteLater)
     thread.finished.connect(thread.deleteLater)
 
     if _connect:
--- a/src/superqt/utils/_qthreading.py
+++ b/src/superqt/utils/_qthreading.py
@@ -881,7 +881,6 @@ def new_worker_qthread(
     worker.moveToThread(thread)
     thread.started.connect(worker.work)
     worker.finished.connect(thread.quit)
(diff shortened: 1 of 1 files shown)


andy-sweet:
The problem with the current change is that the `Worker` is never deleted (unless the caller does it). As it doesn't have a parent, and there no other objects on the new thread as candidate parents, then we can't rely on Qt's object hierarchy to handle it.

My suggestion was to replace the line you deleted with 

```python
thread.finished.connect(worker.deleteLater)
```

which is consistent with [the Qt worker thread example](https://doc.qt.io/qt-6/qthread.html#details).

By making the connection above, the Qt example binds the lifetime of the `Controller` to the execution of the thread and thus the lifetime of the worker - i.e. the client of the `Controller` is responsible for its deletion and that will trigger everything else's (though I don't understand where the `QThread` is deleted...).

By contrast, this superqt function is more automatic and doesn't require the caller to delete anything. The downside of this approach is that as soon as the worker finishes, the calling thread cannot safely refer to the `Worker` instance. As this function also connects `Worker.finished` to `QThread.quit` and `QThread.finished` to `QThread.deleteLater`, the calling thread also cannot safely refer to the `QThread` instance after that time either. I made [a rough diagram of the current code to express potential ordering](https://github.com/andy-sweet/diagrams/blob/main/superqt-threads.drawio.svg).

In summary, I think my suggestion doesn't actually help here - it just delays the deletion of the worker.

But a docstring update could be useful to warn the caller about using the returned `Worker` and `QThread` instances since their lifetimes are hard to predict.

Issue #220, opened 2023-11-22
When update range in `QLargeIntSpinBox` the value is not clipped to new range


Czaki:
**Describe the bug**

When update range in `QLargeIntSpinBox` the value is not clipped to new range

https://github.com/napari/napari/pull/6479#issuecomment-1822181081

**To Reproduce**

```
spin = QLargeIntSpinBox()
spin.setRange(0, 1000)
spin.setValue(500)
spin.setRange(0, 100)
```

and value remains 500 instead of 100

**Expected behavior**

Presented data should be clipped to the new range

cc @jni

tlambert03:
sounds like an easy fix.  anyone want to take this one?  if not, I should be able to get to it next week

Czaki:
I'm not sure where I will have time. But if I got time to write this I will post a message here.

psobolewskiPhD:
Would you just have `setRange` compare `value` to the range, and if outside it then call `setValue`? If that is on the general right track, then I can give it a whirl.

jni:
I think that's the case @psobolewskiPhD. I wouldn't even compare, I'd just do, in `setRange`, `setValue(clip(value, minval, maxval))` for a suitable definition of clip. ;)

Czaki:
> Would you just have `setRange` compare `value` to the range, and if outside it then call `setValue`? If that is on the general right track, then I can give it a whirl.

I suggest put this logic in `setMinimum` and `setMaximum` as this function are called in `setRange`

https://github.com/pyapp-kit/superqt/blob/b927159f49104a8cb90f205d4b82d7ed89a7e1cc/src/superqt/spinbox/_intspin.py#L66-L77

psobolewskiPhD:
> it could be nice to add check if minimum is lower than maximum in setRange

So this is the case where someone does setRange(100, 0) or similar. 
In this case is it better to raise a ValueError or just swap min and max so they are correct?

Czaki:
> So this is the case where someone does setRange(100, 0) or similar.
> In this case is it better to raise a ValueError or just swap min and max so they are correct?

I think that you should mimic Qt behaviour:

```python
from qtpy.QtWidgets import QSpinBox

w = QSpinBox()

w.setRange(0, 100)

w.setRange(1000, 0)

w.minimum(), w.maximum()
Out[5]: (1000, 1000)
```

PR #221, opened 2023-11-24
Bugfix: Check min max versus current value

psobolewskiPhD:
Closes https://github.com/pyapp-kit/superqt/issues/220

As suggested by Grzegorz here https://github.com/pyapp-kit/superqt/issues/220#issuecomment-1824785806 in this PR I have setMin and setMax check versus the current value and if that is smaller/larger it's reset to min/max.

I also added a test for the issue that fails on current main, but passes with this fix.

### DIFF
--- a/src/superqt/spinbox/_intspin.py
+++ b/src/superqt/spinbox/_intspin.py
@@ -65,12 +65,16 @@ class QLargeIntSpinBox(QAbstractSpinBox):
 
     def setMinimum(self, min):
         self._minimum = int(min)
+        if self._minimum > self._value:
+            self.setValue(self._minimum)
 
     def maximum(self):
         return self._maximum
 
     def setMaximum(self, max):
         self._maximum = int(max)
+        if self._maximum < self._value:
+            self.setValue(self._maximum)
 
     def setRange(self, minimum, maximum):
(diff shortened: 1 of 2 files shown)


psobolewskiPhD:
@Czaki also suggested:
> it could be nice to add check if minimum is lower than maximum in setRange

So this is the case where someone does setRange(100, 0) or similar.
In this case is it better to raise a ValueError or just swap min and max so they are correct?

goanpeca:
Let's do what Qt does for it

https://doc.qt.io/qt-6/qabstractslider.html#setRange

>If max is smaller than min, min becomes the only legal value.

Juste teste with QSlider and there seems to be no error, but the widget becomes invisible, I guess because min. and max become the same.

Swapping would be unexpected behavior so I suggest we either follow what Qt does, (puts min and max equal to min?) or raise a value error.

I would prefer the later, not sure what others think :)

psobolewskiPhD:
I was also leaning towards ValueError.
Can then use `with pytest.raises(ValueError):` to test.

Czaki (approved):

tlambert03 (approved):
fantastic.  thanks @psobolewskiPhD!

PR #226, opened 2023-12-15
Use scientific notation for big values in labeled slider

### LINKED ISSUE #224: Use scientific notification for Labeled sliders values for big/small values
In https://github.com/napari/napari/issues/6522 napari got information that for really small numbers the the labels are unreadable:

obraz

The author of Issue suggests that maybe replacing use of scientific notation for representing small values. It may be also useful for large numbers. 

This may be not so trivial as it will require to re implement `SliderLabel` that may require to use `QLabel` as base class. 
 
https://github.com/pyapp-kit/superqt/blob/ef876856267ba869daf02dcd590d2d0e01059ec9/src/superqt/sliders/_labeled.py#L575

What did you think?

Czaki:
Closes #224 

Use `QLabel` instead of `QDoubleSpinBox` for custom representation of big/small values.


tlambert03:
basic idea looks good. lemme know when ready for a full review.  How do you plan to manage when it switches into scientific notation?

Czaki:
Currently, it depends on str(float), as for big values, on my machine it uses scientific notation. I need to check how universal it is.

tlambert03:
hey @Czaki, gentle ping on this PR.  still interested in adding this?

Czaki:
Yes. But I'm not sure if I will find time before January (Christmas etc).

tlambert03:
K, no hurry.  Enjoy

jiban:
Hi @Czaki,
Thank you for the implementation.
By the way, I've looked for an explicit specification that stated `str(float)` used scientific notation when necessary (i.e [*General Format*, 'g'](https://docs.python.org/3/library/string.html#format-specification-mini-language))
I could not find that statement but at least I found that *cpython* implementation was working like that.
`str(float)` uses *General Format* by default.

Here are what I found.
I hope it helps you save time.

1. From [Python Doc. for `str()`](https://docs.python.org/3/library/stdtypes.html#str)
    > str(object) returns [`type(object).__str__(object)`](https://docs.python.org/3/reference/datamodel.html#object.__str__), which is the “informal” or nicely printable string representation of object. For string objects, this is the string itself. If object does not have a [`__str__()`](https://docs.python.org/3/reference/datamodel.html#object.__str__) method, then [`str()`](https://docs.python.org/3/library/stdtypes.html#str) falls back to returning [`repr(object)`](https://docs.python.org/3/library/functions.html#repr).

    so `str(float)` will call `float.__str__()` or `float.__repr__()`
2. Python *float* does not have `__str__()` but have `__repr()__`. See [`PyTypeObject PyFloat_Type`](https://github.com/python/cpython/blob/fb5d1c923677e7982360bad934d70cf9ad3366ca/Objects/floatobject.c#L1889).
3. [`float_repr()`](https://github.com/python/cpython/blob/fb5d1c923677e7982360bad934d70cf9ad3366ca/Objects/floatobject.c#L371) which `float.__repr__` is mapped to calls `PyOS_double_to_string()` with *format_code* of `'r'`.
    ```Python
    static PyObject *
    float_repr(PyFloatObject *v)
    {
        PyObject *result;
        char *buf;

        buf = PyOS_double_to_string(PyFloat_AS_DOUBLE(v),
                                    'r', 0,
                                    Py_DTSF_ADD_DOT_0,
                                    NULL);
        if (!buf)
            return PyErr_NoMemory();
        result = _PyUnicode_FromASCII(buf, strlen(buf));
        PyMem_Free(buf);
        return result;
    }
    ```
4. Finally, [`PyOS_double_to_string()`](https://github.com/python/cpython/blob/fb5d1c923677e7982360bad934d70cf9ad3366ca/Python/pystrtod.c#L787) uses `'g'`([*General Format*](https://docs.python.org/3/library/string.html#format-specification-mini-language)) option for `'r'` *format code*.
    ```Python
    case 'r':          /* repr format */
    /* Supplied precision is unused, must be 0. */
    if (precision != 0) {
        PyErr_BadInternalCall();
        return NULL;
    }
    /* The repr() precision (17 significant decimal digits) is the
       minimal number that is guaranteed to have enough precision
       so that if the number is read back in the exact same binary
       value is recreated.  This is true for IEEE floating point
       by design, and also happens to work for all other modern
       hardware. */
    precision = 17;
    format_code = 'g';
    break;
    ```
so, I think your choice of using `str(float)` is good.

Thank you.

-- Jake

Czaki:
@tlambert03 may you take a look on the implementation of `setValue` to tell what did you think about the logic to converting to string? 

Maybe the minimum and maximum limits should be more flexible:
https://github.com/pyapp-kit/superqt/blob/358d041c0d590850d3e6e89024369f3a4d3f64d1/src/superqt/sliders/_labeled.py#L679-L683

tlambert03 (commented):
> may you take a look on the implementation of setValue to tell what did you think about the logic to converting to string?

seems reasonable to me i think?  Maybe a few more tests would help clarify whether it's doing what you want?


> Maybe the minimum and maximum limits should be more flexible:

sure, seems reasonable

Czaki:
Did I need to add more tests before a review?

tlambert03:
nope, sorry, I missed the request mark.

tlambert03 (approved):
thanks!

src/superqt/sliders/_labeled.py
```diff
@@ -732,21 +817,19 @@ def _update_size(self, *_: Any) -> None:
...
                 w += 8  # it seems as thought suffix() is not enough
         else:
-            w = max(0, _fm_width(fm, self.textFromValue(self.value()))) + 3
+            w = max(0, _fm_width(fm, str(self.value()))) + 3
 
         w += 3  # cursor blinking space
         # get the final size hint
         opt = QStyleOptionSpinBox()
-        self.initStyleOption(opt)
+        # self.initStyleOption(opt)
```
psobolewskiPhD, line 832: @Czaki Why was this change made?
Could it be related to https://github.com/napari/napari/pull/7355#issuecomment-3134232094

tlambert03:
just a note that this commit is also causing issues on ndv, I'm getting consistent segfaults after this commit, but not on the commit before.  https://github.com/pyapp-kit/ndv/issues/213

haven't had time to dig into it yet

PR #246, opened 2024-05-27
Prevent computing full document content highlight per block and only compute current block content for performance

dalthviz:
Related with https://github.com/napari/napari/issues/6845

Current code computes the full document highlight for each document block/paragraph causing freezes as soon the document content starts to get a little bit long. To reproduce, you can run the `examples/code_highlight.py`, copy and paste the widget text content multiple times (selecting the full content to do a new paste increases the chance to stumble with a freeze in a shorter amount of time). So something like:

highlight

This PR tries to prevent/improve the situation by only computing the current block content/text highlight data (not sure if there is a better approach so happy to give a check to any other ideas!). Doing a similar check with the example content copy and paste operations with the changes in this PR:

highlight_perf

### DIFF
--- a/src/superqt/utils/_code_syntax_highlight.py
+++ b/src/superqt/utils/_code_syntax_highlight.py
@@ -1,5 +1,3 @@
-from itertools import takewhile
-
 from pygments import highlight
 from pygments.formatter import Formatter
 from pygments.lexers import find_lexer_class, get_lexer_by_name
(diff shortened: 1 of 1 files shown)


tlambert03:
@Czaki, since this is your object, would appreciate your review

Czaki:
This change provides a significant performance boost and I do not see downsides of these changes. 

Test failures look unrelated.

Czaki (approved):

PR #248, opened 2024-06-03
fix: fix issues with SliderLabel

tlambert03:
I found a case where SliderLabel prevented cleanup by holding onto a bound method.  This cleans that up

### DIFF
--- a/src/superqt/sliders/_labeled.py
+++ b/src/superqt/sliders/_labeled.py
@@ -2,8 +2,7 @@ from __future__ import annotations
 
 import contextlib
 from enum import IntEnum, IntFlag, auto
-from functools import partial
-from typing import Any, Iterable, overload
+from typing import Any, Iterable, cast, overload
 
 from qtpy import QtGui
 from qtpy.QtCore import Property, QPoint, QSize, Qt, Signal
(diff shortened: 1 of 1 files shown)


tlambert03:
this is a real strange mess.  The goal is to remove the `partial` and `lambda` usage in _labeled... but for some reason, merely commenting out the line:
```
        if connect is not None:
            self.editingFinished.connect(lambda: connect(self.value()))
```
in SliderLabel causes a segfault on pyside6.

not urgent, so going back in the pile for now

Czaki:
@tlambert03 could you point line in code?

tlambert03:
Sure, basically the goal of this PR is to:

- remove use of partial [here](https://github.com/pyapp-kit/superqt/blob/952ac336bf7555101f3ce4b622417d0a17057732/src/superqt/sliders/_labeled.py#L574)
- get rid of connect argument [here](https://github.com/pyapp-kit/superqt/blob/952ac336bf7555101f3ce4b622417d0a17057732/src/superqt/sliders/_labeled.py#L645) and the connection that happens in the init [here](https://github.com/pyapp-kit/superqt/blob/952ac336bf7555101f3ce4b622417d0a17057732/src/superqt/sliders/_labeled.py#L665-L666) ... instead, having anything that instantiates a SliderLabel worry about doing its own connections.
-in doing that last step, get rid of the [lambda here](https://github.com/pyapp-kit/superqt/blob/952ac336bf7555101f3ce4b622417d0a17057732/src/superqt/sliders/_labeled.py#L666)

Czaki:
Ah. You do not use `tox`/`nox` that makes it harder to reproduce locally. 

Could you try to check if store hard reference to `connect` when commenting out this part of code?  
Because commenting out lambda connection means also lost the reference.

tlambert03:
prefer not to use nox/tox honestly.  
however, i don't think it should be hard to reproduce. i just made a new environment, pip installed editable with tests, and installed the pyside6 extra (==6.7) and was able to reproduce it.  i'm on mac fwiw.

don't have time to check the references now

Czaki:
I'm going sleep. If you do not do this, I will try tomorrow.

PR #251, opened 2024-06-15
feat: graceful offline fallback for qiconify

tlambert03:
QIconifyIcon requires internet connection for the first load (after which it is cached and reused).  However, it still shouldn't fail when offline, so this PR has it fall back to a question mark with a warning, allowing the program to continue loading


alisterburt:
🫡

PR #256, opened 2024-10-15
feat: categorical slider [WIP]

tlambert03:
This adds a new `QCategoricalSlider` and `QLabeledCategoricalSlider`, when you want a slider to select from a discrete set of objects (rather than numbers).  Similar to a combobox I guess, but easier to change between all the values.

API still needs consideration

https://github.com/user-attachments/assets/43ab9ead-2282-45c2-b166-79975e37648a


Czaki:
Maybe also add an enum version of it?

PR #260, opened 2024-11-26
fix: minimum size hint for QElidingLabel

### LINKED ISSUE #259: QElidingLabel hesitates to decrease size within layouts
**Describe the bug**
It seems that when `QElidingLabel` shares a layout, it...hesitates...to decrease its size. You have to release the mouse button, and click the edge of the window again, to get the width to decrease to the next "sticking point".

I don't really know what's going on, so the best I can do is provide a MCVE and a gif (below). It seems to be an issue with the lack of a minimum size hint - with my limited understanding of [layout management](https://doc.qt.io/qt-5/layout.html#adding-widgets-to-a-layout), I threw the following WIPpy method into `QElidingLabel`, which seemingly fixes it :

```python
    def minimumSizeHint(self) -> QSize:
        fm = QFontMetrics(self.font())
        flags = int(self.alignment() | Qt.TextFlag.TextWordWrap)
        if not self.wordWrap():
            r = fm.boundingRect(QRect(QPoint(0, 0), self.size()), flags, "...")
            return QSize(r.width(), r.height())
        r = fm.boundingRect(QRect(QPoint(0, 0), self.size()), flags, "...")
        return QSize(self.width(), r.height())
```

**To Reproduce**
I see this behavior on `main`, by running the following script:
```python

from qtpy.QtWidgets import QWidget, QPushButton, QHBoxLayout
from superqt import QElidingLabel

app = QApplication([])


class Wdg(QWidget):
    def __init__(self):
        super().__init__()
        self._widget = QElidingLabel(
            "a skj skjfskfj sdlf sdfl sdlfk jsdf sdlkf jdsf dslfksdl sdlfk sdf sdl "
            "fjsdlf kjsdlfk laskdfsal as lsdfjdsl kfjdslf asfd dslkjfldskf sdlkfj"
        )
        btn = QPushButton("Btn")
        layout = QHBoxLayout(self)
        layout.addWidget(btn)
        layout.addWidget(self._widget)

    def resizeEvent(self, a0):
        print(self._widget.minimumSizeHint())
        return super().resizeEvent(a0)
# widget.setWordWrap(True)
widget = Wdg()
widget.show()
app.exec_()
```

**Expected behavior**
I would expect that, when you decrease the size of a widget containing multiple children, the `QElidingLabel` would be...willing(?) to decrease its size down to some minimal text. I don't know whether that should just be "...", or maybe the first letter followed by "...".
**Screenshots**
Without my function fix:
Code_LuPuGH4czY
With the function:
Code_hQHd77MPSy

**Desktop (please complete the following information):**
 - Windows 11
 - PyQt5
 - Python 3.11.0

gselzer:
Closes #259

### DIFF
--- a/src/superqt/elidable/_eliding_label.py
+++ b/src/superqt/elidable/_eliding_label.py
@@ -73,3 +73,10 @@ class QElidingLabel(_GenericEliding, QLabel):
         flags = int(self.alignment() | Qt.TextFlag.TextWordWrap)
         r = fm.boundingRect(QRect(QPoint(0, 0), self.size()), flags, self._text)
         return QSize(self.width(), r.height())
+
+    def minimumSizeHint(self) -> QSize:
+        # The smallest that self._elidedText can be is just the ellipsis.
+        fm = QFontMetrics(self.font())
+        flags = int(self.alignment() | Qt.TextFlag.TextWordWrap)
+        r = fm.boundingRect(QRect(QPoint(0, 0), self.size()), flags, "...")
+        return QSize(r.width(), r.height())
--- a/tests/test_eliding_label.py
+++ b/tests/test_eliding_label.py
@@ -69,3 +69,14 @@ def test_wrap_text():
     assert isinstance(wrap, list)
     assert all(isinstance(x, str) for x in wrap)
     assert 9 <= len(wrap) <= 13
+
+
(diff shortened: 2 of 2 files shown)


tlambert03 (approved):
Lovely

PR #268, opened 2024-12-23
feat: Improve CodeSyntaxHighlight object

tlambert03:
This PR improves a number of things and fixes some slight issues with `CodeSyntaxHighlight`:

- adds full type hinting, including `KnownStyle` for IDE suggestions of valid pygments names
- prevents Formatter from overriding all font selections on the parent object.  I tried to apply fonts and found that `get_text_char_format` was always clobbering whatever I did.  Furthermore, it wasn't successfully applying "Monospace", instead I get: `WARNING: Populating font family aliases took 75 ms. Replace uses of missing font family "Monospace" with one that exists to avoid this cost.`.  So this adds a full font family tree that should work on any system, *and* it only applies it if the style explicitly sets the "mono" flag.  This means that most users will now need to set Monospace themselves (but since it already wasn't working on, macos for example, this is probably a good thing).
- allows the user to pass in any QObject that supports `document()`, simplifying the usage, and also allowing allowing us to automatically set the background color on objects that support setPalette.
- adds `setTheme` and `setLanguage` methods that allow you to change the style on an existing formatter, including updating the background color on supported objects.
- adds lots of docstrings to explain what's going on

### DIFF
--- a/src/superqt/utils/_code_syntax_highlight.py
+++ b/src/superqt/utils/_code_syntax_highlight.py
@@ -231,6 +231,7 @@ class CodeSyntaxHighlight(QSyntaxHighlighter):
         self.set_theme(theme)
 
     def set_theme(self, theme: KnownStyle | str) -> None:
+        """Set the theme for the syntax highlighting."""
         self.formatter = QFormatter(style=theme)
         if self._doc_parent is not None:
             palette = self._doc_parent.palette()
--- a/src/superqt/utils/_code_syntax_highlight.py
+++ b/src/superqt/utils/_code_syntax_highlight.py
@@ -1,81 +1,256 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any, cast
+
+import pygments.token
 from pygments import highlight
 from pygments.formatter import Formatter
 from pygments.lexers import find_lexer_class, get_lexer_by_name
 from pygments.util import ClassNotFound
-from qtpy import QtGui
+from qtpy.QtGui import (
+    QColor,
+    QFont,
+    QPalette,
+    QSyntaxHighlighter,
+    QTextCharFormat,
+    QTextDocument,
+)
+
+if TYPE_CHECKING:
+    from collections.abc import Mapping, Sequence
+    from typing import Literal, TypeAlias, TypedDict, Unpack
+
+    import pygments.style
+    from pygments.style import _StyleDict
+    from pygments.token import _TokenType
+    from qtpy.QtCore import QObject
+
+    class SupportsDocumentAndPalette(QObject):
+        def document(self) -> QTextDocument | None: ...
+        def palette(self) -> QPalette: ...
+        def setPalette(self, palette: QPalette) -> None: ...
+
+    KnownStyle: TypeAlias = Literal[
+        "abap",
+        "algol",
+        "algol_nu",
+        "arduino",
+        "autumn",
+        "bw",
+        "borland",
+        "coffee",
+        "colorful",
+        "default",
+        "dracula",
+        "emacs",
+        "friendly_grayscale",
+        "friendly",
+        "fruity",
+        "github-dark",
+        "gruvbox-dark",
+        "gruvbox-light",
+        "igor",
+        "inkpot",
+        "lightbulb",
+        "lilypond",
+        "lovelace",
+        "manni",
+        "material",
+        "monokai",
+        "murphy",
+        "native",
+        "nord-darker",
+        "nord",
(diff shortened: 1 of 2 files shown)


tlambert03:
I assume the two failing napari tests are unrelated, but @Czaki could you confirm?

src/superqt/utils/_code_syntax_highlight.py
```diff
@@ -1,81 +1,267 @@
...
-        text_char_format.setBackground(QtGui.QColor(style["bgcolor"]))
+MONO_FAMILIES = [
+    "Menlo",
+    "Courier New",
+    "Courier",
+    "Monaco",
+    "Consolas",
+    "Andale Mono",
+    "Source Code Pro",
+    "monospace",
```
Czaki, line 100: I think that it may be worth to and `Ubuntu Mono` - the default mono font on Ubuntu system.

src/superqt/utils/_code_syntax_highlight.py
```diff
@@ -1,81 +1,267 @@
...
+    from pygments.style import _StyleDict
+    from pygments.token import _TokenType
+    from qtpy.QtCore import QObject
+
+    class SupportsDocumentAndPalette(QObject):
+        def document(self) -> QTextDocument | None: ...
+        def palette(self) -> QPalette: ...
+        def setPalette(self, palette: QPalette) -> None: ...
+
+    KnownStyle: TypeAlias = Literal[
```
Czaki, line 32: any plan for trace pigments style updates (ex compare pygments.styles.STYLES content witrh this annotation in tests) or just wait for issues?

src/superqt/utils/_code_syntax_highlight.py
```diff
@@ -1,81 +1,267 @@
...
+    - ansicolor: str | None
+    - bgansicolor: str | None
+    """
+    text_char_format = QTextCharFormat()
+    if style.get("mono"):
+        text_char_format.setFontFamilies(MONO_FAMILIES)
+    if color := style.get("color"):
+        text_char_format.setForeground(QColor(f"#{color}"))
+    if bgcolor := style.get("bgcolor"):
+        text_char_format.setBackground(QColor(bgcolor))
```
Czaki, line 129: Why one with hash, and the other without?

Czaki (approved):
looks good

src/superqt/utils/_code_syntax_highlight.py
tlambert03: good point!  it was already that way (on main):

https://github.com/pyapp-kit/superqt/blob/4da5ac262cb1197713b77fd3f776aea17f4236f1/src/superqt/utils/_code_syntax_highlight.py#L20-L24

you wrote that ... do you recall any reason for it?

src/superqt/utils/_code_syntax_highlight.py
tlambert03: no plan... but I'll mention that the type hint here is actually:

```python
theme: KnownStyle | str = "default",
```

which means that users will get the autocomplete suggestions, but it's not an error to enter something that is not in the KnownStyle list.  So the only "issue" we might see is that someone didn't get a nice autocomplete hint :)

tlambert03:
> you do not set up windows manager https://pytest-qt.readthedocs.io/en/latest/troubleshooting.html#xvfb-assertionerror-timeouterror-when-using-waituntil-waitexposed-and-ui-events

i do use https://github.com/pyvista/setup-headless-display-action ... which replaces all three of `tlambert03/setup-qt-libs@v1`, `powershell gl-ci-helpers/appveyor/install_opengl.ps1`, and `aganders3/headless-gui@v2`.  You can see it run here: https://github.com/pyapp-kit/superqt/actions/runs/12485275920/job/34843848003?pr=268

the only thing I think they don't use is `herbstluftwm` ... does that make all the difference here?

> Maybe we should think to also easy skip test marked as skip_local_focus on other CI.

that would be useful.  some sort of mark rather than a local skip.  will find some workaround for now.  did you also have to change your CI setup in napari when that change went in?  I don't remember

Czaki:
> the only thing I think they don't use is `herbstluftwm` ... does that make all the difference here?

Yes. It is required to propagate GUI events.

>  did you also have to change your CI setup in napari when that change went in

We have added `herbstluftwm`, but it is added to `aganders3/headless-gui@v2`

tlambert03:
TIL.  Thanks

Added it to setup headless action in https://github.com/pyvista/setup-headless-display-action/pull/25 (and that fixes it here too)

PR #270, opened 2024-12-31
Lazy-import `pyconify`

hanjinliu:
I found that ~50% of the import time of `superqt` originates from `pyconify` even though only `QIconifyIcon` depends on it.
This PR moves the import statements inside the methods.

In my local environment, the execution time of `import superqt` was:
main: 0.864 s (pyconify: 0.445 s)
this PR: 0.436 s

### DIFF
--- a/src/superqt/__init__.py
+++ b/src/superqt/__init__.py
@@ -11,7 +11,6 @@ except PackageNotFoundError:
 from .collapsible import QCollapsible
 from .combobox import QColorComboBox, QEnumComboBox, QSearchableComboBox
 from .elidable import QElidingLabel, QElidingLineEdit
-from .iconify import QIconifyIcon
 from .selection import QSearchableListWidget, QSearchableTreeWidget
 from .sliders import (
     QDoubleRangeSlider,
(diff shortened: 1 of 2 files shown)


tlambert03:
Good catch!  This is interesting, i wonder if we can fix pyconify more directly to be faster to import?Do you have any sense for what is slowing it down?  On Dec 31, 2024, at 2:42 AM, codecov[bot] ***@***.***> wrote:﻿
Codecov Report
All modified and coverable lines are covered by tests ✅

Project coverage is 87.16%. Comparing base (6a7a731) to head (48af284).

Additional details and impacted files
@@           Coverage Diff           @@
##             main     #270   +/-   ##
=======================================
  Coverage   87.15%   87.16%           
=======================================
  Files          46       46           
  Lines        3465     3466    +1     
=======================================
+ Hits         3020     3021    +1     
  Misses        445      445           

☔ View full report in Codecov by Sentry.
📢 Have feedback on the report? Share it here.

hanjinliu:
According to the report, most (>90%) of the import time came from `request`, so maybe delaying the import of `request` in `pyconify` is better.

tlambert03:
Yeah that would be great.  Didn't expect requests would be slow to load, but since pyconify will usually use a local cached version, we should delay importing requests until necessary (ie only when internet needed)

tlambert03:
ok, requests is now delayed in https://github.com/pyapp-kit/pyconify/pull/27

if we want to do something here as well, I'd prefer we follow the pattern in the other optional modules, moving the iconify import to this part of `__init__`

https://github.com/pyapp-kit/superqt/blob/6a7a731c5d67e388a826445e4a5d3fc77f953e73/src/superqt/__init__.py#L53-L67

tlambert03:
done in [1a7e8e0](https://github.com/pyapp-kit/superqt/pull/270/commits/1a7e8e0c429f8f17c293f2cdecd48120b30017fe)

PR #275, opened 2025-01-26
Update CONTRIBUTING.md to install .[dev] first then pre-commit

psobolewskiPhD:
Small change.
Need to have `pre-commit` installed before `pre-commit install` works.


tlambert03:
thanks

Issue #274, opened 2025-01-26
Add text filtering to QColormapComboBox


psobolewskiPhD:
CmapCatalogComboBox has really nice filtering using the LineEdit, which QColormapComboBox doesn't have.

Obviously the Catalog is extensive so it makes sense there, but a list of colormaps provided to QColormapComboBox can also be long, so just typing to filter can be really nice.

I can't think of use cases where the LineEdit filtering would be bad?
But maybe a `QColormapFilterComboBox` is the way to go, to keep the simpler approach for very short lists?

tlambert03:
i can see that being useful.  happy to see a PR if you have time

PR #278, opened 2025-01-30
Enh: Adds a filterable kwarg to ColormapComboBox enabling filtering (like Catalog)

psobolewskiPhD:
Closes https://github.com/pyapp-kit/superqt/issues/274

In this PR I add a `filterable` kwarg to QColormapComboBox to add the filtering that the CmapCatalog one has.
I more or less copied the approach from CmapCatalog with the addition of the model (with the help of Copilot) to make it work. Plus, I modified `_PopupColormapLineEdit` so that click opens the combobox, like QColormapComboBox, but you can still highlight the text or double click (effectively triple click) to select it, so you can type to do the filtering.
I modified an existing test to also test both cases of filterable.
Feedback very welcome.


psobolewskiPhD:
We should see the napari tests pass!

tlambert03:
Thanks for this and sorry for the delay @psobolewskiPhD.  I like the general functionality here.  I think my only concern is the possible confusion surrounding another top level widget that subclasses an existing widget with *very* similar functionality to that already-existing one, but a slightly different name.  Would it be at all possible to add this functionality to the existing widget, gated by a parameter `__init__(filterable: bool)`?  Essentially... just make your new widget *the* `QColormapComboBox`, and guard the additional functionality you've added here behind `setEditable()` (or whatever property you think is most relevant to the new functionality).  I know it will make a slightly more complicated single class, but I think the benefit of that is that the public API becomes less confusing

psobolewskiPhD:
Oh that totally makes sense! I should have thought of that!
I will give it a shot, maybe later this week.
❤️

psobolewskiPhD:
@tlambert03 with your tip it turned out to be a pretty easy refactor and it's quite clean I think!
Definitely an improvement.
Only think I'm not sure on is if the handling of `self.setFilterable(False)` is sufficient.
I did local testing and it works as expected, but maybe I also need to disable more things?

tlambert03:
thanks @psobolewskiPhD, this is great.  I think we might as well make this the default behavior.  I'm going to play with it for a moment to see if i can slightly improve the behavior when typing something that ends up with no hits. (It would be nice if it were impossible to leave the combobox in an invalid state).

psobolewskiPhD:
@tlambert03 I find that if you enter an invalid name it just shows the ? and the insertion bar keeps blinking. I can't actually exit it? (escape, enter)
image

Also, when testing, i hit this unexpected surprise:
if you enter a name from the catalog (that is not in the list provided to the combobox), it works!
image

tlambert03:
Yeah I noticed both of those too... fixing now to make it always show a valid camp, and to add sneakily entered cmaps to the list

tlambert03:
ok, I made some changes. I

- make filterable default behavior
- simplify setFilterable logic (don't recreate the completer and all that stuff, just toggle it's use)
- add allow_invalid option to line edit, which is false by default, so you can't leave the line edit in an invalid state
- add manually typed, correct values to the dropdown (so you know the name you're looking for you can quickly add it)

tlambert03:
can you take it for a test drive and see if it still works for you?

psobolewskiPhD:
@tlambert03 Works great!
you're a wizard 😄

PR #280, opened 2025-02-20
Fix themes compatibility for QRangeSlider

Jordan-Pierce:
Add custom styles for `QRangeSlider` to ensure compatibility with qdarkstyle.

* Update the `setStyleSheet` method in `src/superqt/sliders/_generic_range_slider.py` to include the custom styles for qdarkstyle.
* Update the `update_styles_from_stylesheet` function in `src/superqt/sliders/_range_style.py` to handle stylesheets more effectively.
* Add documentation in `docs/widgets/qrangeslider.md` to guide users on how to apply custom styles for compatibility with qdarkstyle.

### DIFF
--- a/docs/widgets/qrangeslider.md
+++ b/docs/widgets/qrangeslider.md
@@ -227,3 +227,31 @@ setSliderPosition(val: Sequence[int]) -> None
 ```python
 sliderMoved(Tuple[int, ...])
 ```
+
+## Custom Styles for Compatibility with qdarkstyle
+
+To ensure compatibility with qdarkstyle, you can add custom styles for `QRangeSlider` as shown below:
+
+```python
+QSS = """
+QSlider {
+    background-color: none;
+}
+
+QSlider::add-page:vertical {
+    background: none;
+    border: none;
+}
+
+QRangeSlider {
+    qproperty-barColor: #9FCBFF;
+}
+"""
+```
+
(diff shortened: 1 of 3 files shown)


Jordan-Pierce:
@tlambert03 going to stop trying to make edits for now and leave it be if you want to take a gander. Sorry for all the commits.

tlambert03 (commented):
thanks for working on this @Jordan-Pierce!  Definitely appreciate it.  I know this is a frustrating topic, and an inelegant current situation.

Unfortunately, this solution isn't *quite* right.  Before this PR, on macos, it looks just like the system theme:

Screenshot 2025-02-21 at 1 06 49 PM

After this PR, the hard-coded barColor will only work in certain situations, and no longer matches system theme on macos (and likely others).  And also, the bar is now offset (i know that's confusing, since you didn't manually adjust it... but that's the state of things)

Screenshot 2025-02-21 at 1 07 01 PM

This sort of os-specific issue is why that `_range_style.py` module is so complex to begin with.  So I think any fix will likely need to go in that module, rather than a static set of css that gets unconditionally applied.  (sorry it's so annoying)

Jordan-Pierce:
@tlambert03 understood, I'll keep trying to chip away at it.

tlambert03:
thanks! I really do appreciate it!

Issue #249, opened 2024-06-03
Breaking Refactor: Make GenericSliders inherit `QWidget` rather than `QSlider` or `QAbstractSlider`


tlambert03:
The original design of all the Range sliders here (which all subclass `sliders._generic_slider._GenericSlider`) was to make them behave as much like drop-in replacements for `QSlider` as possible.  I wanted them all to pass `isinstance(obj, QSlider)` checks, and inasmuch as possible, inherit all of the behaviors of `QSlider`. 

However, that has proven difficult to maintain, and the biggest issue relates to signals (particularly with PySide).  In most of the slider variants the `value()` (and therefore `valueChanged` signal) is of a different type than the superclass.  `QDoubleSlider`, of course, has a `value()` of type `float`, while `QRangeSlider` has a value type of `tuple[int, ...]`.  

We do some "unsavory" patching of the signal instances, on the object instances themselves, like [here](https://github.com/pyapp-kit/superqt/blob/main/src/superqt/sliders/_generic_slider.py#L93-L96) and [here](https://github.com/pyapp-kit/superqt/blob/main/src/superqt/sliders/_generic_range_slider.py#L42-L43).  And, for the most part it works, but requires very ugly hacks (https://github.com/pyapp-kit/superqt/pull/51).  But really, it doesn't **work well**... and it's just a matter of time before it breaks again.  I already get misc bus errors and segfaults during tests on pyside6 with the current strategy (thought pyqt6 is fine).

So, the two main options that I can see are:

1. Don't try to inherit from `QAbstractSlider`, since it already has the signal names we want to use, and changing the value type of signals has proven to be a bad idea.  A refactor like this would mean either using composition, or re-implementing basically all of the methods in the C implementation of QAbstractSlider to make these objects essentially duck-QSliders (many of those are reimplemented anyway as is).  They would no longer pass `isinstance()`, but they would still be mostly drop-in replacements.
2. Use new (different) names signals that change types.  For example `QDoubleSlider.valueChanged` would be `QDoubleSlider.floatValueChanged` or something like that, `rangeChanged` would have to be `floatRangeChanged`... and `QRangeSlider.valueChanged` might be `QRangeSlider.valuesChanged`, etc...  This breaks a lot of existing code, but *might* be cleaner to implement.

this doesn't have to happen soon, but it's a bit of a ticking time bomb.  @Czaki and @jni, just a heads up if you have any opinions.

jni:
[...]

PR #283, opened 2025-03-16
fix: less Slider signal renaming, make alternate signal types public

tlambert03:
related to https://github.com/pyapp-kit/superqt/issues/249 ... the signals on range and labeled sliders have proven to be problematic.  The original goal was to make the api be as close as possible to Qt sliders, so that you could use a DoubleSlider or a RangeSlider as a drop-in replacement, with names like `valueChanged`, etc...  But the need to essentially override the parent class signals with a new signature was probably a bad idea, and has proven problematic, particularly for cleanup.

I'm currently seeing lots of cleanup segfaults over in [ndv](https://github.com/pyapp-kit/ndv), and have narrowed it down to these signals.  In my particular case, I'm using `QLabeledRangeSlider` and `QLabeledSlider` (not the float variants), and lines like `self.valueChanged = self._ivalueChanged` in `QLabeledSlider._rename_signals` were what was causing the segfaults, mostly on pyside.

This PR does two things:
- avoids overwriting signals when the new signal has the exact same signature.  For example `QLabeledSlider._ivalueChanged = Signal(int)`  already had the same signature as the parent `QAbstractSlider.valueChanged`, and was therefore unnecessary (explicit and clear perhaps, but unnecessary).   Those are now just removed.k
- makes public signals that have a different type.  for example, `QRangeSlider._valuesChanged` is now public at `QRangeSlider .valuesChanged`.  This is a *first* step towards #249.  It simply makes available the other signal for connection, without deprecating or removing the (problematic) `QRangeSlider.valueChanged`  alias.  This could create a path towards deprecation, allowing libraries to use different signals before we remove the problematic ones

cc @Czaki.  a quick review would be appreciated.  This shouldn't affect anything for you, but just want you to be aware

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ ci:
 
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.8.6
+    rev: v0.9.4
     hooks:
       - id: ruff
         args: [--fix, --unsafe-fixes]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ ci:
 
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.9.4
+    rev: v0.9.9
     hooks:
       - id: ruff
         args: [--fix, --unsafe-fixes]
@@ -17,7 +17,7 @@ repos:
       - id: validate-pyproject
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.14.1
+    rev: v1.15.0
     hooks:
       - id: mypy
         exclude: tests|examples
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,7 +12,7 @@ To get started fork this repository, and clone your fork:
 git clone https://github.com/<your_organization>/superqt
 cd superqt
 
-# install in editable mode
+# install in editable mode (this will install PyQt6 as the Qt backend)
 pip install -e .[dev]
 
 # install pre-commit hooks
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,6 +65,7 @@ dev = [
     "pydocstyle",
     "rich",
     "types-Pygments",
+    "superqt[test,pyqt6]",
 ]
 docs = [
     "mkdocs-macros-plugin ==1.3.7",
--- a/src/superqt/__init__.py
+++ b/src/superqt/__init__.py
@@ -56,9 +56,9 @@ __all__ = [
 ]
 
 if TYPE_CHECKING:
-    from .combobox import QColormapComboBox  # noqa: TC004
-    from .iconify import QIconifyIcon  # noqa: TC004
-    from .spinbox._quantity import QQuantity  # noqa: TC004
+    from .combobox import QColormapComboBox
+    from .iconify import QIconifyIcon
+    from .spinbox._quantity import QQuantity
 
 
 def __getattr__(name: str) -> Any:
--- a/src/superqt/combobox/__init__.py
+++ b/src/superqt/combobox/__init__.py
@@ -13,7 +13,7 @@ __all__ = (
 
 
 if TYPE_CHECKING:
-    from superqt.cmap import QColormapComboBox  # noqa: TC004
+    from superqt.cmap import QColormapComboBox
 
 
 def __getattr__(name: str) -> Any:  # pragma: no cover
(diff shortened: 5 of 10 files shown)


tlambert03:
looks like this does specifically break something in the napari tests,  that we apparently don't have covered here.  so need to look at the failing napari tests

tlambert03:
napari tests were failing because https://github.com/napari/napari/pull/5757 added a test that [specifically checks for the presence of the private attribute `_valuesChanged`](https://github.com/napari/napari/blob/cb6f6e6157990806a53f1c58e31e9e7aa4a4966e/napari/_qt/layer_controls/_tests/test_qt_layer_controls.py#L336).  I suppose using `isinstance()` would be better there... but in any case, I just updated the code here to emit a warning when the (now missing) private attribute is accessed

jni:
Thanks for investigating @tlambert03!

Just to make sure I understand cos I haven't looked at that code before: you recommend changing `if getattr(qslider, '_valuesChanged', None):` to `if isinstance(qslider, SUPERQT_SLIDER_CLASSES):`? (For a suitable definition of that list?)

Reading in your description above, "This is a first step towards Breaking Refactor: Make GenericSliders inherit QWidget rather than QSlider or QAbstractSlider", it makes me think that the [loop above that if statement](https://github.com/napari/napari/blame/cb6f6e6157990806a53f1c58e31e9e7aa4a4966e/napari/_qt/layer_controls/_tests/test_qt_layer_controls.py#L334) also needs to be changed: `for qslider in ctrl.findChildren(QAbstractSlider):` would no longer find superqt sliders? Any suggestions for fixing that? 😬 I guess just loop over *all* QObjects and then filter by QAbstractSlider *and* the superqt classes?

tlambert03:
> you recommend changing if getattr(qslider, '_valuesChanged', None): to if isinstance(qslider, SUPERQT_SLIDER_CLASSES):? (For a suitable definition of that list?)

i suppose, I didn't really look too closely at the test.  I wasn't entirely sure what was going on, but it looks like it's using the presence of an attribute as a sort of duck-typing to determine whether it's a range slider.  it says the comment `"in the case the slider is ranged"` ... so I think I would indeed just check `isinstance(slider, superqt.QRangeSlider)` rather than `if getattr(qslider, '_valuesChanged', None):` ...   (the `_valuesChanged` attribute doesn't seem to be used once checked, so it really does appear to be a sort of duck-typing isinstance check.


> it makes me think that the [loop above that if statement](https://github.com/napari/napari/blame/cb6f6e6157990806a53f1c58e31e9e7aa4a4966e/napari/_qt/layer_controls/_tests/test_qt_layer_controls.py#L334) also needs to be changed: for qslider in ctrl.findChildren(QAbstractSlider): would no longer find superqt sliders

yeah, that's true from the title of that old PR... but I wouldn't worry about it yet.  Note that option number two from [that pr](https://github.com/pyapp-kit/superqt/issues/249) would retain inheriting from AbstractSlider

> Use new (different) names signals that change types. For example QDoubleSlider.valueChanged would be QDoubleSlider.floatValueChanged or something like that, rangeChanged would have to be floatRangeChanged... and QRangeSlider.valueChanged might be QRangeSlider.valuesChanged, etc... This breaks a lot of existing code, but might be cleaner to implement.

so really, this is a step towards that one, and i don't think you should be worried about `ctrl.findChildren(QAbstractSlider)` for now.  That would be a major breaking change (if it ever happens) and that one won't happen without a clear deprecation

jni:
> it really does appear to be a sort of duck-typing isinstance check.

yes that was my interpretation also. Ok PR incoming. 😊

tlambert03:
thanks!

tlambert03:
I'm going to just leave the `_valuesChanged` in there as an alias for valuesChanged.  looks like that test was in v0.4.19 as well... so if we want to continue to test that changes to superqt don't break backwards compatibility for older versions of napari, i either need to skip that test specifically, or just leave the attribute in there.  It's easy enough to just leave it in there

Czaki:
I was busy this weekend. But when review after merge it looks good.

tlambert03:
thank you for looking after merge.  sorry I didn't wait (but I'm always happy to make post-merge changes if you find anything)

jni:
blech. Sorry I missed it.

I think it's not a big deal to break the tests of older napari versions, as long as the functionality isn't broken. It would be difficult to get an environment working with v0.4.19 and superqt-latest anyway. 😅

jni:
Fyi, napari/napari#7719. But as you say, no rush.

tlambert03:
> think it's not a big deal to break the tests of older napari versions, as long as the functionality isn't broken. It would be difficult to get an environment working with v0.4.19 and superqt-latest anyway. 😅

That’s actually what we are currently testing routinely.  It does work.  Happy to stop testing older versions, but I think @Czaki wanted to have those tests, so I’ll leave them active unless he agrees

tlambert03:
Thanks @Czaki 

Ultimately we’ll have to leave the private attribute in there for napari until we only test at least napari 7719 and later, so it’s not urgent

Issue #287, opened 2025-04-05
test_cmap_draw_result fails on big endian (s390x)


penguinpee:
**Describe the bug**
I first observed `test_cmap_draw_result` failing on big endian in version 0.7.2. When I updated the Fedora package to 0.7.3 I happened to hit a big endian builder on which that same test failed again.

```python
=================================== FAILURES ===================================
____________________________ test_cmap_draw_result _____________________________
    def test_cmap_draw_result():
        """Test that the image drawn actually looks correct."""
        # draw into any QPaintDevice
        w = 100
        h = 20
        pix = QPixmap(w, h)
        cmap = Colormap("viridis")
        draw_colormap(pix, cmap)
    
        ary1 = cmap(np.tile(np.linspace(0, 1, w), (h, 1)), bytes=True)
        ary2 = qimage_to_array(pix.toImage())
    
        # there are some subtle differences between how qimage draws and how
        # cmap draws, so we can't assert that the arrays are exactly equal.
        # they are visually indistinguishable, and numbers are close within 4 (/255) values
        # and linux, for some reason, is a bit more different``
        atol = 8 if platform.system() == "Linux" else 4
>       np.testing.assert_allclose(ary1, ary2, atol=atol)
tests/test_cmap.py:64: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
args = (<function assert_allclose.<locals>.compare at 0x3ff5bfe4b80>, array([[[ 68,   1,  84, 255],
        [ 68,   3,  87, 2...
        ...,
        [231, 247, 255,  24],
        [231, 255, 255,  33],
        [231, 255, 255,  33]]], dtype=uint8))
kwds = {'equal_nan': True, 'err_msg': '', 'header': 'Not equal to tolerance rtol=1e-07, atol=8', 'verbose': True}
    @wraps(func)
    def inner(*args, **kwds):
        with self._recreate_cm():
>           return func(*args, **kwds)
E           AssertionError: 
E           Not equal to tolerance rtol=1e-07, atol=8
E           
E           Mismatched elements: 7720 / 8000 (96.5%)
E           Max absolute difference: 128
E           Max relative difference: 16.
E            x: array([[[ 68,   1,  84, 255],
E                   [ 68,   3,  87, 255],
E                   [ 69,   8,  91, 255],...
[...]

PR #288, opened 2025-04-06
Make qimage_to_array() work on big endian

penguinpee:
Make sure the returned ndarray is ordered the same as on little
endian systems.

Solves #287

### DIFF
--- a/src/superqt/utils/_img_utils.py
+++ b/src/superqt/utils/_img_utils.py
@@ -1,3 +1,5 @@
+import sys
+
 from typing import TYPE_CHECKING
 
 from qtpy.QtGui import QImage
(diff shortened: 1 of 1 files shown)


src/superqt/utils/_img_utils.py
```diff
@@ -37,4 +38,8 @@ def qimage_to_array(img: QImage) -> "np.ndarray":
     arr = np.frombuffer(b, np.uint8).reshape(h, w, c)
 
     # reverse channel colors for numpy
-    return arr.take([2, 1, 0, 3], axis=2)
+    # On big endian we need to specify a different order
+    if sys.byteorder == "big":
+        return arr.take([1, 2, 3, 0], axis=2)
```
Czaki, line 43: Should we add here  `  # pregma: no cover` to mark as not expected to be covered?

penguinpee: Should I amend this or will you?

Czaki, line 43:
```suggestion
        return arr.take([1, 2, 3, 0], axis=2)  # pragma: no cover
```

PR #290, opened 2025-04-18
fix: Set SliderProxy range params to Any

gselzer:
When the types are `int`s, this raises mypy errors when e.g. setting the range of a `QDoubleSlider` to float values.

This change aligns with the parameter typing of `_SliderProxy.setValue`

### DIFF
--- a/src/superqt/sliders/_labeled.py
+++ b/src/superqt/sliders/_labeled.py
@@ -83,7 +83,7 @@ class _SliderProxy:
     def setPageStep(self, step: int) -> None:
         self._slider.setPageStep(step)
 
-    def setRange(self, min: int, max: int) -> None:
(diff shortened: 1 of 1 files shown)


tlambert03:
Would float satisfy mypy? (I think it works for int too)

src/superqt/sliders/_labeled.py
```diff
@@ -83,7 +83,7 @@ def pageStep(self) -> int:
     def setPageStep(self, step: int) -> None:
         self._slider.setPageStep(step)
 
-    def setRange(self, min: int, max: int) -> None:
+    def setRange(self, min: Any, max: Any) -> None:
```
tlambert03, line 86:
```suggestion
    def setRange(self, min: float, max: float) -> None:
```

PR #304, opened 2025-08-12
fix: Fix size of Slider label to not ocupy to much space

Czaki:
Follow up to #226. 
Solves problem pointed out by @psobolewskiPhD  https://github.com/napari/napari/pull/8184#issuecomment-3141341450

I found that, even if parent is set, the qss rules are not applied. 

I found that recalculation of size may be triggered on showEvent. 

Other option is to check for `StyleChange` event in `changeEvent` method

### DIFF
--- a/src/superqt/sliders/_labeled.py
+++ b/src/superqt/sliders/_labeled.py
@@ -6,7 +6,11 @@ from typing import TYPE_CHECKING, Any, overload
 
 from qtpy import QtGui
 from qtpy.QtCore import Property, QPoint, QSize, Qt, Signal
-from qtpy.QtGui import QDoubleValidator, QFontMetrics, QValidator
+from qtpy.QtGui import (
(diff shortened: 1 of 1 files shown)


src/superqt/sliders/_labeled.py
```diff
@@ -6,15 +6,19 @@
 
 from qtpy import QtGui
 from qtpy.QtCore import Property, QPoint, QSize, Qt, Signal
-from qtpy.QtGui import QDoubleValidator, QFontMetrics, QValidator
+from qtpy.QtGui import (
+    QDoubleValidator,
+    QFontMetrics,
+    QValidator,
+)
```
tlambert03, line 13:
```suggestion
from qtpy.QtGui import QDoubleValidator, QFontMetrics, QValidator
```

tlambert03 (approved):
works for me if it does what you need

PR #320, opened 2026-01-02
Fix values of slider label when range/value is set outside default

TimMonko:
I noticed that in the camera popup widget in napari that the value of the label in slider labels no longer shows value outside the default range of the sliders (0,99). https://github.com/napari/napari/issues/8534 (Should the decimal issue, which I proposed to fix in napari like this: https://github.com/napari/napari/pull/8535/files#diff-b7be858b0c42314c0f7fbaaa128a2e6a24822a61aa780d9374153c02a7a6b30f actually be fixed in superqt?)

I added a test here that all assertions fail on main, but pass with this PR.

I don't know if this code change is the correct fix because this is my first time looking inside superqt, but my thought process was that the label range needs to be connected to changing the slider range. 

I'm only confused because this isn't how it used to work (see the screenshot in https://github.com/napari/napari/pull/7626), but a quick scan of the file history isn't making anything stand out in this method as changed, so maybe this isn't the right place to fix this / the cause of the bug. It seems both the range and decimals somehow got disconnected from the label

### DIFF
--- a/tests/zz_test_sliders/test_labeled_slider.py
+++ b/tests/zz_test_sliders/test_labeled_slider.py
@@ -99,8 +99,8 @@ def test_slider_label_value(qtbot):
 
     slider.setValue(-90)
     assert slider.value() == -90 == slider._label.value()
-    
-    
+
+
 def test_slider_label_decimals_update_text(qtbot):
     slider = QLabeledDoubleSlider()
     qtbot.addWidget(slider)


tlambert03:
Hi @TimMonko 

If something has mysteriously changed at some point, and it's not clear where the fix should go, would you be able to run git bisect with superqt installed in editable mode and narrow down exactly what superqt commit caused the change?  That might help us fill in the picture a bit and make sure we're fixing the right thing for you

Czaki:
I was worried that it was that commit...

Czaki (approved):
For me, these changes are ok. I might work on allowing the selection of slider label type (QLineEdit or QDoubleSpinBox) base in a separate PR.

DragaDoncila:
@tlambert03 would we be able to get a release of superqt soon? We'd love to have these improvements to the slider in napari 0.7.0

PR #321, opened 2026-01-02
chore: Group dependabot updates in a single PR

Czaki:
To avoid the creation of multiple PR with updates of actions, just configure dependabot to open a single PR.


tlambert03:
Awesome! Didn't know that was possible now

PR #325, opened 2026-01-30
fix: Use faint color for disabled on-state toggle switch

hanjinliu:
In the current implementation, we use the same toggle swtich groove color for enabled+off, disabled+on and disabled+off. This makes the check state unclear when the widget is disabled.

In this PR, on-color and off-color are mixed so it is visually easier to find the on-state switches, and they do not stand out too much. 

Examples of enabled vs disabled toggle switches:

image


tlambert03 (approved):
Fine by me! Thanks

tlambert03 (approved):
Fine by me! Thanks

PR #334, opened 2026-03-29
refactor: use style-sheet free QToolButton for QCollapsible

tlambert03:
This removes the hard-coded (fragile, non-theme-aware) stylesheet from the QCollapsible.  It shouldn't change the default style for most use cases and applications (unless those applications are already manually styling all widgets, in which case things *may* change), but it "plays better" with the existing app.

@Czaki @jni @DragaDoncila ... the *one* place you will notice this in napari is in the napari-plugin-manager, which uses this.

currently for you, it looks like this:

Screenshot 2026-03-29 at 12 23 43 PM

the toggle button *has* a background.  (That's actually a bit of an "accident" anyway, because `superqt.QCollapsible` attempted to render as background-free with a stylesheet of:

```python
self._toggle_btn.setStyleSheet("text-align: left; border: none; outline: none;")
```

so I think the fact that you see a background is due to your global application stylesheet applying a background to all QPushButtons, somehow overriding the superqt one here (or else, it's being done by the napari-plugin-manager styles somewhere).

With this PR, your plugin manager will look like this:

Screenshot 2026-03-29 at 12 22 52 PM

... and that's because this PR changes it from a QPushButton to a QToolButton (to be able to use `setAutoRaise(True)`... to make it look like a ghost button, which was the original intention of the `border: none; outline: none;` stylesheet hack that was previously applied.

Can you guys live with that change?  Or, if not, update your stylesheets to change it how you want it to appear?

### DIFF
--- a/src/superqt/collapsible/_collapsible.py
+++ b/src/superqt/collapsible/_collapsible.py
@@ -15,6 +15,7 @@ from qtpy.QtCore import (
 )
 from qtpy.QtGui import QIcon, QPainter, QPaintEvent, QPalette, QPixmap
 from qtpy.QtWidgets import (
+    QApplication,
     QFrame,
     QSizePolicy,
     QStyle,
@@ -33,6 +34,8 @@ class _GhostToolButton(QToolButton):
         super().__init__(parent)
         if title:
             self.setText(title)
+        # Match QPushButton typography from main for consistent header text size.
+        self.setFont(QApplication.font("QPushButton"))
         self.setCheckable(True)
         self.setAutoRaise(True)
         self.setToolButtonStyle(Qt.ToolButtonStyle.ToolButtonTextBesideIcon)
--- a/src/superqt/collapsible/_collapsible.py
+++ b/src/superqt/collapsible/_collapsible.py
@@ -13,8 +13,39 @@ from qtpy.QtCore import (
     Qt,
     Signal,
 )
-from qtpy.QtGui import QIcon, QPainter, QPalette, QPixmap
-from qtpy.QtWidgets import QFrame, QPushButton, QSizePolicy, QVBoxLayout, QWidget
+from qtpy.QtGui import QIcon, QPainter, QPaintEvent, QPalette, QPixmap
+from qtpy.QtWidgets import (
+    QFrame,
+    QSizePolicy,
+    QStyle,
+    QStyleOptionToolButton,
+    QStylePainter,
+    QToolButton,
+    QVBoxLayout,
+    QWidget,
+)
+
+
+class _GhostToolButton(QToolButton):
+    """Tool button that keeps a ghost appearance while checked."""
+
+    def __init__(self, parent: QWidget | None = None, *, title: str = "") -> None:
+        super().__init__(parent)
+        if title:
+            self.setText(title)
+        self.setCheckable(True)
+        self.setAutoRaise(True)
+        self.setToolButtonStyle(Qt.ToolButtonStyle.ToolButtonTextBesideIcon)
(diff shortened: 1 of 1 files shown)


src/superqt/collapsible/_collapsible.py
```diff
@@ -13,8 +13,41 @@
...
+        self.setToolButtonStyle(Qt.ToolButtonStyle.ToolButtonTextBesideIcon)
+
+    def paintEvent(self, a0: QPaintEvent | None) -> None:
+        if not a0:
+            return super().paintEvent(a0)
+        option = QStyleOptionToolButton()
+        self.initStyleOption(option)
+        option.state &= ~QStyle.StateFlag.State_On
+        painter = QStylePainter(self)
+        painter.drawControl(QStyle.ControlElement.CE_ToolButtonLabel, option)
```
Czaki, line 50: I'm not sure how to fix this, but this branch prevents from apply any style using stylesheet

Czaki (commented):
Hi, Sorry for long time to response, but debug this was not trivial. 

So what I found, that current implementation of paintEvent prevent us from fix the look using the style sheet. It looks like your code ignores at least background part of style sheet when painting. 

I have opened a PR to napari-plugin-manager that contains fixes in stylesheet to style this button https://github.com/napari/napari-plugin-manager/pull/210

You mentioned that you opened this PR to use [autoRaise](https://doc.qt.io/qt-6/qtoolbutton.html#autoRaise-prop) and from description it should draw a frame around 

> QToolButton supports auto-raising. In auto-raise mode, the button draws a 3D frame only when the mouse points at it. The feature is automatically turned on when a button is used inside a [QToolBar](https://doc.qt.io/qt-6/qtoolbar.html). Change it with [setAutoRaise](https://doc.qt.io/qt-6/qtoolbutton.html#autoRaise-prop)(). 

Are you sure that current paintEvent implementation do not prevent it? In napari this effect is not visible. 

The last problem is vertical alignment. It is visible on your screenshots, but when enable styling, it is more visible. 

Image

As I tested, a potential napari user will not get any feedback that Installation Info is clickable, as neither stylesheet nor auto raise give mouse feedback. 

So from my point of view this PR require some polishing, and I'm not sure if not easier will be to subclass QPushButton and add hover effect using stylesheet or in paint event.

tlambert03:
ok, thanks @Czaki.  I'll try to look into why stylesheets might not being applied sometime soon

tlambert03:
ok, there are workarounds:

```python
  def paintEvent(self, a0: QPaintEvent | None) -> None:                                                          
      option = QStyleOptionToolButton()                                                                          
      self.initStyleOption(option)                                                                               
      option.state &= ~(QStyle.StateFlag.State_On | QStyle.StateFlag.State_Sunken)                               
      painter = QStylePainter(self)                                                                              
      interactive = option.state & (                                                                             
          QStyle.StateFlag.State_MouseOver | QStyle.StateFlag.State_HasFocus                                     
      )                                                                                                          
      has_sheet = self.style().metaObject().className() == "QStyleSheetStyle"                                    
      if interactive or has_sheet:                                                                               
          painter.drawPrimitive(QStyle.PrimitiveElement.PE_PanelButtonTool, option)                              
      painter.drawControl(QStyle.ControlElement.CE_ToolButtonLabel, option)  
```

but if you're keen to keep the exact same look as the original version that had stylesheets directly applied, more experimentation is probably needed... I don't have time for that at the moment, so i can also just deal with this in by overriding own usage wherever I use this component.

Czaki:
Could we just make a button class, an argument of constructor of QCollapsible?

Issue #340, opened 2026-05-06
QToggleButton animation de-synced when sate changed inside QSignalBlocker


carlosmariorr:
**Describe the bug**
The QToggleSwitch class won't update the GUI if used under a QSignalBlocker.

This is happening because the animation is handled as an event connection in:
```python 
class QToggleSwitch(QtW.QAbstractButton):
    
    ....

    def __init__(  # type: ignore [misc]  # overload
        self, text: str | None = None, parent: QtW.QWidget | None = None
    ) -> None:
        
        ....

        self.toggled.connect(self._animate_handle)
```

**To Reproduce**
1. Create a QToggleSwitch.
2. Update the state inside a QSignalBlocker.
3. The button will glitch, properly changing the color but not the GUI handle.  

**Expected behavior**
If the button state is updated inside a QSignalBlocker, the GUI should be updated alongside the state.

**Screenshots**
Button state `Checked = True` but GUI handle `OFF` (WRONG state)
<img width="212" height="55" alt="Image" src="https://github.com/user-attachments/assets/b372a9ee-68c7-4114-a0c4-bab3d81c7742" />

Button state `Checked = False` but GUI handle `ON` (WRONG state)
<img width="248" height="51" alt="Image" src="https://github.com/user-attachments/assets/8660f415-bd81-4700-b917-a4ce43b879b9" />

Button state `Checked = False` GUI handle `OFF` (This is CORRECT)
<img width="240" height="50" alt="Image" src="https://github.com/user-attachments/assets/f6a705b0-65d9-4d0d-80ca-13916eec0376" />

**Desktop:**
 - Windows 11 (26200.8328)
 - PyQt6 (6.10.2)
 - Python 3.13

Czaki:
@carlosmariorr Are you sure that your screenshot's description is correct?

carlosmariorr:
Yes @Czaki , when the button checked state is False, the color is gray and handle should be on the left.  I added a correct screenshot for comparison. 

When the button checked state is True, the color is blue and the handle should be on the right. 

This is how it looks like if I toggle in napari-metadata with the `QSignalBlocker`:

<img width="426" height="240" alt="Image" src="https://github.com/user-attachments/assets/1b63f756-f233-45b9-91c2-d50e0a0df8fc" />

PR #341, opened 2026-05-06
fix: Sync the toggle switch state when `setChecked()` is run inside a `QSignalBlocker` 

carlosmariorr:
## Relevant Issues:
#340 

## Issue solved:
It was not properly syncing before whenever the `setChecked()` method was called within a `QSignalBlocker`. This is creating a bug in `napari-metadata` as shown in #340.

 The current problem is that the animation is connected to the `toggle` event and it will be eaten by the `QSignalBlocker` so the color changes but the animation never runs and thus the handle is de-synced. 
 
 This is the best way I could think of syncing back without a somewhat bigger refactor. 

## Changes:
- Added an override to the `checkStateSet()` function that was not being used in `QToggleSwitch` to sync the state of the `isChecked()` method with the toggle button's handle (In the GUI). 
- Added tests to check if the handle is not synced with the button's state. 

Maybe @tlambert03 or @Czaki have a better solution?

### DIFF
--- a/src/superqt/switch/_toggle_switch.py
+++ b/src/superqt/switch/_toggle_switch.py
@@ -115,7 +115,7 @@ class QToggleSwitch(QtW.QAbstractButton):
         width = opt.switch_width + text_size.width() + opt.margin * 2 + 8
         return QtCore.QSize(width, height)
 
-    # This fixes the animation when QSignalBlocker is used. It is hacky.
+    # This fixes the animation when QSignalBlocker is used.
     def checkStateSet(self) -> None:
         """Do animation if signals are blocked."""
         super().checkStateSet()
--- a/src/superqt/switch/_toggle_switch.py
+++ b/src/superqt/switch/_toggle_switch.py
@@ -51,7 +51,6 @@ class QToggleSwitch(QtW.QAbstractButton):
         super().__init__(parent)
         self.setCheckable(True)
         self.setCursor(Qt.CursorShape.PointingHandCursor)
-        self.toggled.connect(self._animate_handle)
 
         self._anim = QtCore.QPropertyAnimation(self, b"_offset", self)
         self._anim.setDuration(120)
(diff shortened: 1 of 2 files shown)


Czaki:
I think that the line `self.toggled.connect(self._animate_handle)` should be removed and everything should be implemented in `checkStateSet` without checking of blocked signal.

carlosmariorr:
I asked codex to make a sync function to handle the states since programatic changes go through `checkStateSet()` but mouse clicks go through `nextCheckState()`. 

Its working right now on my side with no connections to `toggle` event. I don't know if this will break other stuff but I couldn't find anything else that doesn't go through `nextCheckState()` or `checkStateSet()`.

src/superqt/switch/_toggle_switch.py
```diff
@@ -274,10 +283,14 @@ def _set_handleSize(self, size: int) -> None:
...
 
+    def _sync_handle_position(self) -> None:
+        self._anim.stop()
+        self._animate_handle(self.isChecked())
+
     def _animate_handle(self, val: bool) -> None:
         end = self._offset_for_checkstate(val)
         if self._anim.duration():
-            self._anim.setStartValue(self._offset_for_checkstate(not val))
+            self._anim.setStartValue(self._offset_value)
```
carlosmariorr, line 293: I changed this because the previous version made it a little jumpy since it always starts from the opposite side. Was that intended?

src/superqt/switch/_toggle_switch.py
```diff
@@ -274,10 +283,14 @@ def _set_handleSize(self, size: int) -> None:
 
     ### Other private methods ###
 
+    def _sync_handle_position(self) -> None:
+        self._anim.stop()
+        self._animate_handle(self.isChecked())
+
```
carlosmariorr, line 289: Stopping the animation just in case and then setting the handle based on the state. This function is required to connect it to both `checkStateSet()` and `nextCheckState()`.

tests/test_toggle_switch.py
```diff
@@ -142,3 +142,59 @@ def test_multiple_lines(qtbot):
...
+    assert wdg._offset == off
+    assert not wdg.isChecked()
+
+    wdg.setChecked(True)
+    assert wdg.isChecked()
+    assert wdg._offset == on
+
+    wdg.setChecked(False)
+    assert not wdg.isChecked()
+    assert wdg._offset == off
```
carlosmariorr, line 165: I think this might be testing too much since the sync is no longer connected to the `toggle` event. Should I get rid of this or should I keep it to have both, the `signals blocked` and the `no signals blocked` checks?

carlosmariorr (commented):
I think this works instead of connecting to the `toggle` event. It got rid of the bugs I had on `napari-metadata` and I tested the buttons on a separate widget I made and other buttons in the `features table` of napari. All three were working fine.

tlambert03:
@hanjinliu ... i think this one was your feature right?  Would you mind reviewing this?

src/superqt/switch/_toggle_switch.py
hanjinliu: I think this is because `_animate_handle` was used for `toggled` signal, where start value should be the opposite of the incoming value.

tests/test_toggle_switch.py
hanjinliu: let's keep both. It doesn't take much time either way.

hanjinliu (approved):
I tested locally and this looks good to me. Thanks, @carlosmariorr !

tlambert03 (approved):
thanks @carlosmariorr!

PR #342, opened 2026-05-07
fix: disallow QQuantity with derived units

### LINKED ISSUE #339: `QQuantity` does not work for derrived units
**Describe the bug**
The unit field of a `QQuantity` will not show the correct conversions for derived units, e.g. Newtons, or meters per second. Instead, it will show attempted conversions for the first base unit in the derived unit, e.g. kilograms. If one attempts to switch to another unit, it will raise an error.

**To Reproduce**
```python
import sys

from PyQt6.QtWidgets import QApplication
from superqt import QQuantity


def main() -> int:
    app = QApplication.instance() or QApplication(sys.argv)
    quant = QQuantity("25 N")
    quant.show()
    return app.exec()


if __name__ == "__main__":
    raise SystemExit(main())
```

**Expected behavior**
This could easily get out of hand fast. For `N`, I would think other common units might be `kgf`, `lbf`, though I'm not sure I really want to suggest using a different unit systems / Pint Contexts.

I know that inverse units aren't officially supported right now (in comparison to derived units, they don't seem so bad), so maybe it would be better to specify derived units also aren't supported.

Relevant lines:
https://github.com/pyapp-kit/superqt/blob/3285493e93e3813f43384a2a4e382033850dbd94/src/superqt/spinbox/_quantity.py#L124-L131

**Screenshots**

Image

**Desktop (please complete the following information):**
 - Windows 11
 - Python 3.13.9
 - superqt 0.8.1

Antyos:
Fixes #339.

**Added:**
- Units with exponents (`1/s`, `m^2`) are supported
- Raises a `NotImplementedError` when using non-simple units (`m/s`, `N`).

There is probably a world where it wouldn't be too hard to infer the prefixes if it's a single unit like `N`, but I'm not tackling that right now. 

I have also updated the docs and added corresponding tests.

### DIFF
--- a/docs/widgets/qquantity.md
+++ b/docs/widgets/qquantity.md
@@ -30,8 +30,8 @@ app.exec()
 
 !!! note
 
-    `QQuantity` currently support simple units with exponents, e.g., `meters^2` or
-    `1/second`. However, non-simple units, e.g. `meter/second`, `Newton`, etc.,
+    `QQuantity` currently supports simple units with exponents, e.g., `meters^2` or
+    `1/second`. However, compound units, e.g. `meter/second`, `Newton`, etc.,
     are not currently supported.
 
 {{ show_widget(150) }}
--- a/docs/widgets/qquantity.md
+++ b/docs/widgets/qquantity.md
@@ -28,6 +28,12 @@ w.show()
 app.exec()
 ```
 
+!!! note
+
+    `QQuantity` currently support simple units with exponents, e.g., `meters^2` or
+    `1/second`. However, non-simple units, e.g. `meter/second`, `Newton`, etc.,
+    are not currently supported.
+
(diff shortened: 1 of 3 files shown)


docs/widgets/qquantity.md
```diff
@@ -28,6 +28,12 @@ w.show()
 app.exec()
 ```
 
+!!! note
+
+    `QQuantity` currently support simple units with exponents, e.g., `meters^2` or
+    `1/second`. However, non-simple units, e.g. `meter/second`, `Newton`, etc.,
```
tlambert03:
```suggestion
    `QQuantity` currently supports simple units with exponents, e.g., `meters^2` or
    `1/second`. However, compound units, e.g. `meter/second`, `Newton`, etc.,
```

PR #343, opened 2026-05-07
chore: Update compound unit error message

Antyos:
I updated the error message in the code to reflect the "non-simple" -> "compound" wording you fixed in the docs in #342.


tlambert03:
thanks!

PR #344, opened 2026-05-11
chore: Update napari test workflows

TimMonko:
napari/napari#8704

1. bump python version to 3.13 for more recent (we will drop py 3.10 in the near future)
2. uses `dependency-group` from workflow setup

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -103,7 +103,7 @@ jobs:
     secrets: inherit
 
   test_napari:
-    uses: pyapp-kit/workflows/.github/workflows/test-dependents.yml@v2
+    uses: pyapp-kit/workflows/.github/workflows/test-dependents.yml@65abcc0b091a0d45dbe3c589266160cc9b62f8dc
     with:
       dependency-repo: napari/napari
       dependency-group: "testing"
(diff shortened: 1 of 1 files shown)


TimMonko:
@tlambert03 would love to fix the actions here, but I can't find any report on why the startup is failing. I tried the conventional commit fix like I did in https://github.com/pyapp-kit/app-model/pull/285, but that doesn't seem to work. Maybe the runners are just down? 🍻

TimMonko:
Wait, nevermind I can read...
https://github.com/pyapp-kit/superqt/actions/runs/25692867011?pr=344

tlambert03:
yeah, just change the pin from `@v2` to `@main`

.github/workflows/test_and_deploy.yml
```diff
@@ -103,13 +103,13 @@ jobs:
     secrets: inherit
 
   test_napari:
-    uses: pyapp-kit/workflows/.github/workflows/test-dependents.yml@v2
+    uses: pyapp-kit/workflows/.github/workflows/test-dependents.yml@65abcc0b091a0d45dbe3c589266160cc9b62f8dc
```
TimMonko:
```suggestion
    uses: pyapp-kit/workflows/.github/workflows/test-dependents.yml@main
```

PR #346, opened 2026-05-13
Reuse value format logic for label width determination in sliders

### LINKED ISSUE #335: QLabeledDoubleRangeSlider label width too small
**Describe the bug**
The width calculation logic in `SliderLabel` appears to be not very robust for use withing `QLabeledDoubleRangeSlider`.
E.g. when setting the range using integer values, all label fields (min, max, and values) are cut off (see Example 1 below). When setting the range using floats, the label width is slightly bigger, but still not big enough, and importantly it doesn't scale properly when increasing the number of decimals (see Example 2).

**To Reproduce**
***Example 1:***
```python
from PyQt5 import QtWidgets as QtW
from PyQt5.QtCore import Qt
from superqt import QLabeledDoubleRangeSlider

app = QtW.QApplication([])

slider = QLabeledDoubleRangeSlider(Qt.Orientation.Horizontal)
slider.setRange(-1, 1)
slider.setValue((-0.5, 0.5))
slider.show()

app.exec_()
```

***Example 2:***
```python
from PyQt5 import QtWidgets as QtW
from PyQt5.QtCore import Qt
from superqt import QLabeledDoubleRangeSlider

app = QtW.QApplication([])

slider = QLabeledDoubleRangeSlider(Qt.Orientation.Horizontal)
slider.setRange(-1.0, 1.0)
slider.setDecimals(5)
slider.setValue((-0.5, 0.5))
slider.show()

app.exec_()
```

**Expected behavior**
The slider labels should grow to fit the entire value.

**Screenshots**
***Example 1:***
Image
The labels only show the numbers after the decimal point, but upon focusing them it is possible to scroll through to see they indeed contain the full value.

***Example 2:***
Image

**Desktop:**
 - OS: Windows 11 Version 10.0.26200 Build 26200
 - Qt Backend: PyQt5
 - Python version: 3.13.12

brisvag:
- closes #335

There is still a small overlap between the handle label text if they are both long. Not sure why that's the case. At least it solves the "cropping" problem from the original issue.

Before:
image

After:

image

### DIFF
--- a/src/superqt/sliders/_labeled.py
+++ b/src/superqt/sliders/_labeled.py
@@ -735,8 +735,8 @@ class SliderLabel(QLineEdit):
         self._value = val
         self.updateText()
 
-    def updateText(self) -> None:
-        val = float(self._value)
+    def _format_value(self, value):
+        val = float(value)
         use_scientific = (abs(val) < 0.0001 or abs(val) > 9999999.0) and val != 0.0
         font_metrics = QFontMetrics(self.font())
         eight_len = _fm_width(font_metrics, "8")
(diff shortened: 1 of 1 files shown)


Czaki (approved):

tlambert03:
thanks @brisvag

PR #348, opened 2026-05-18
QQuantity specify compound unit options

Antyos:
It was bugging me that the `QQuantity` was useless for anyone using a compound unit when it seemed like such a simple feature to allow the user to specify the units they want to use. So here we are.

A few things I'd like some thoughts on:

- I went with the term `units_options` to be consistent with the naming of other internal variables, like the `units` parameter and `QQuantity._units_combo`, but I think that `unit_options` makes more grammatical sense.

- I redid some of the type annotations to use `pint.facets.plain.PlainQuantity` and `pint.facets.plain.PlainUnit` because my Pylance was putting red squiggles on a lot of things. For instance, creating a unit out of the `ureg` directly seems to give a `PlainQuantity` rather than a `pint.Quantity` (which inherits from `PlainQuantity`). I probably could've implemented this with better intentionality, but the squiggles are gone. I don't mind reverting some of this if you want.

- `QQuantity.setValue()` allows for changing the dimensionality of the quantity. If so, should it also accept a `units_options` (or `unit_options`) parameter?

- The docs probably need to be updated again.

### DIFF
--- a/src/superqt/spinbox/_quantity.py
+++ b/src/superqt/spinbox/_quantity.py
@@ -2,6 +2,8 @@ from typing import TYPE_CHECKING, TypeAlias, Union
 
 try:
     from pint import Quantity, Unit, UnitRegistry
+    from pint._typing import UnitLike
+    from pint.facets.plain.unit import PlainUnit
     from pint.util import UnitsContainer
 except ImportError as e:
     raise ImportError(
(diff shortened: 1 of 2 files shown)


Antyos:
Well, I'm super confused why the tests aren't working. I don't know enough about Pint to explain this and I can't reproduce it on my machine to test further. I'd guess different unit registries, but they should both be using the same one? No clue.

```
tests\test_quantity.py:67: AssertionError
=========================== short test summary info ===========================
FAILED tests/test_quantity.py::test_qquantity_compound_units - AssertionError: assert <Quantity(3.6, 'kilometer / hour')> == <Quantity(3.6, 'kilometer / hour')>
 +  where <Quantity(3.6, 'kilometer / hour')> = value()
 +    where value = <superqt.spinbox._quantity.QQuantity object at 0x0000019F6BE3DE10>.value
 +  and   <Quantity(3.6, 'kilometer / hour')> = <class 'pint.Quantity'>(3.6, 'km/h')
 +    where <class 'pint.Quantity'> = <pint.registry.UnitRegistry object at 0x0000019F6981D1E0>.Quantity
 +      where <pint.registry.UnitRegistry object at 0x0000019F6981D1E0> = unitRegistry()
 +        where unitRegistry = <superqt.spinbox._quantity.QQuantity object at 0x0000019F6BE3DE10>.unitRegistry
============ 1 failed, 413 passed, 55 skipped, 1 xfailed in 9.02s =============
```

